PR #87, opened 2024-04-11
setup: move client/.env  to client/.env.example

bernhardreiter:
* .. to raise chances that values are inspected and to not have a configuration file that is changed for use in production in the repository.


ThomasJunk (approved):
LGTM

ThomasJunk:
I added the `.env` to `.gitignore` in order to prevent unintended commits of `.env`.

PR #114, opened 2024-04-30
Check user permission

koplas:
Checks if the user can make API requests and redirects to the login page, if not.


ThomasJunk (approved):
LGTM

PR #127, opened 2024-05-10
Add Groups to role creation, add misc. changes that are necessary to modernize scripts

JanHoefelmeyer:


cintek (approved):
@JanHoefelmeyer and I discussed some details of this PR internally and they are changed now. I ran the main script, tested the application, ran the scripts to install/prepare the smaller parts and everything works fine.

PR #132, opened 2024-05-14
Refactor table

koplas:
This reduces code duplication and allows to change the displayed columns dynamically.

### DIFF
--- a/client/src/lib/Advisories/Overview.svelte
+++ b/client/src/lib/Advisories/Overview.svelte
@@ -9,152 +9,29 @@
 -->
 
 <script lang="ts">
-  /* eslint-disable svelte/no-at-html-tags */
   import { onMount } from "svelte";
-  import { appStore } from "$lib/store";
-  import { push } from "svelte-spa-router";
-  import {
-    Button,
-    Label,
-    PaginationItem,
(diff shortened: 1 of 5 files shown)


cintek (changes requested):
Overall, this works and looks good. Thanks. There are only minor issues:

1. `savePosition` can be found very often in the code. I think it would be better to watch `offset`, `currentPage`, `limit` and `orderBy` and when they change automatically call `savePosition`
2. Now, if the user enters a letter in the search bar the application starts a request right away. Here it would be better to use a small delay and then start the request. If the user changes the input within the delay you would reset the "countdown". Then, we would reduce the amount of requests that have to be handled by the server.

koplas:
> If you enter some characters in the search bar so that the number of result becomes smaller and then mark all characters and delete them there is no new request so the list won't be updated.

Done

cintek (approved):
Thanks!

bernhardreiter:
Nitpicking the title: 

> and allows to change the displayed columns dynamically.

makes it something else than a "code refactoring", as this is
"the process of restructuring existing computer code [..] without changing its external behavior"
(cited from wikipedia). >;)

PR #145, opened 2024-05-29
feat: Extend workflowstates

ThomasJunk:
Extend the given workflowstates with the backward paths from the `Admin`.


s-l-teichmann:
No. E.g.  `{ReadWorkflow, DeleteWorkflow}:        {Editor, Reviewer},` Editors should not be allowed to move documents to the delete state. only Reviewers should be able to do so.

ThomasJunk:
My bad, I took it from [main](https://github.com/ISDuBA/ISDuBA/blob/main/pkg/models/workflow.go#L47) without checking it.

client/src/lib/permissions.ts
s-l-teichmann, line 1: I think the matrix in the client should be machine generated to have a the single source of truth.

s-l-teichmann (changes requested):

s-l-teichmann (approved):
From my side this done now. -> LGTM

JanHoefelmeyer (approved):
LGTM & works

Issue #172, opened 2024-06-14
Diff parsing problem


bernhardreiter:
@tschmidtb51 found a parsing or display problem for the side-by-side diff display:

![image](https://github.com/ISDuBA/ISDuBA/assets/8190008/47eb1659-8cfd-4f97-9c34-2ee809cf2a57)

The characters like `[-` and `-]` are probably not in the original files.

To compare, here is the integrated diff for the same section:
![Screenshot 2024-06-13 at 14-48-37 ICSA-23-108-02](https://github.com/ISDuBA/ISDuBA/assets/8190008/77308f83-ca13-4ea8-b469-d0ba8bbf1450)

s-l-teichmann:
The problem may arise from the delimiters `{+`, `+}`,  `[-`  and `-]`.  I've took the idea of choose these symbols from the `--word-diff=plain` mode from Git.

Without further escaping this is not robust against containing delimiters inside a section.

@cintek 
I therefore propose to let a JSON representation do the segmentation part for us. Instead of a single string let us use a list of commands like `[{'m': 'i', 't': 'foo'}, {'m': 'o', 't': 'bar'}, {'m': 'd', 't': 'baz'}]` with `m` for **mode** and `t` for **text**, `i` for **insert**, `d` for **delete** and `o` for **original**.

This is be more verbose, but it should make the parsing easier/more robust and aligns well with the the `add`, `delete` and `insert` sections idea of the JSONPatch stuff.

cintek:
@s-l-teichmann indeed, this looks more robust than the previous solution and is easy to implement in Svelte so I agree to do it this way.

PR #179, opened 2024-06-18
Change word diff to JSON segmentation.

s-l-teichmann:
Solves #172

### DIFF
--- a/client/src/lib/Diff/DiffEntry.svelte
+++ b/client/src/lib/Diff/DiffEntry.svelte
@@ -20,12 +20,14 @@
 </script>
 
 <div style={containerStyle}>
-  {#if Array.isArray(content)}
+  {#if Array.isArray(content) && !content[0]["m"]}
     {#each content as val, index}
       <div class="mb-4 flex">
-        {index + 1}.&ensp;<svelte:self content={val} depth={depth + 1}></svelte:self>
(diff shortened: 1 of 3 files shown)


s-l-teichmann:
@cintek Please add the client part

cintek:
@s-l-teichmann in general it is working but I observed that sometimes I get a JSON object like this:
```
{
  op: "replace",
  path: "/vulnerabilities/0/scores/0/cvss_v3/baseScore",
  value: 7.8
}
```

I saw this two times with the `baseScore`. Everything else seems to work as expected. I thought the problem is that it is a number but on the other hand if `document/tracking/version` or `/document/tracking/generator/engine/version` changed the value of the object is correct.

s-l-teichmann:
> @s-l-teichmann in general it is working but I observed that sometimes I get a JSON object like this:
> 
> ```
> {
>   op: "replace",
>   path: "/vulnerabilities/0/scores/0/cvss_v3/baseScore",
>   value: 7.8
> }
> ```
> 
> I saw this two times with the `baseScore`. Everything else seems to work as expected. I thought the problem is that it is a number but on the other hand if `document/tracking/version` or `/document/tracking/generator/engine/version` changed the value of the object is correct.

This is out of focus for this issue. The word-diff logic is only applied to text fields.

cintek:
@JanHoefelmeyer can you please test the JSON diff component?

JanHoefelmeyer:
LGTM & works

PR #180, opened 2024-06-18
Add initial downloader

### LINKED ISSUE #13: Robin wants to get an overview of all csaf document sources of the system
Robin wants to get an overview of all sources of the system

### LINKED ISSUE #47: Implement Source-Management
Currently, the Source-Page is empty sans the word "Sources". The functionality still needs to be implemented.

### LINKED ISSUE #211: fully implement source-manager
Source managers are responsible for the upkeep of the download of advisories and the management of sources. 

https://github.com/ISDuBA/ISDuBA/pull/180 contains a lot of the work necessary to implement it.

### LINKED ISSUE #212: Fully implement importer
Importers are responsible for the download of advisories. 

https://github.com/ISDuBA/ISDuBA/pull/180 contains a lot of the work necessary to implement it.

koplas:
This pull request adds a job system that manages the import of CSAF documents.

This closes #13, closes #47, closes #211 and closes #212.


JanHoefelmeyer (changes requested):
A few initial things that bothered me:
- [ ] setup scripts no longer work and no documentation on how to fix
- [x] everyone can access sources tab and create (but not save) jobs, should be disabled for everyone but importer/source-manager
- [x] given "start job" and "delete job" are right next to each other, I'd prefer a warning if you want to delete it
- [ ] editing a job resets several values: start and end date, the three checkboxes at the bottom, client key and passphrase unless they are explicitly saved via the browser (which is the wrong way to handle this, for multiple reasons)
- [ ] start and end ranges using MM/DD/YYYY is inconsistent with the downloader using YYYY-MM-DD
- [x] tasks can be clicked, but clicking them doesn't do anything
- [x] all job logs are simply dumped into the repository's main directory. 
- [x] When starting a job, status doesn't update itself on completion (but it does upon refresh, site reload, abortion etc.)
- [x] aborted job will be shown as completed upon refresh
- [ ] schedules cannot be edited. Would expect a similar method to jobs.

- [ ] idea: job name, schedule name and cron expression show example values. These could be pre-entered(, with the name schema: 'Schedule #X', X∈ℕ, X being the smallest nontaken number and the cron expression either being the one used on the last created cron job with the shown expression as default or with a configurable default)

 - [ ] input-fields need to check whether their input is valid. Some try to do so, but aren't perfect (e.g. number of workers only accepts integers, but accepts all from ℤ, not only ℕ), some fail completely and accept any input.

I will keep checking for a bit, but those were the first few things I found.

cintek:
>   everyone can access sources tab and create (but not save) jobs, should be disabled for everyone but importer/source-manager

Fixed.

>   given "start job" and "delete job" are right next to each other, I'd prefer a warning if you want to delete it

Now we show a warning and the icon buttons are a bit larger so it's easier to hit them.

>  start and end ranges using MM/DD/YYYY is inconsistent with the downloader using YYYY-MM-DD

I think this is okay because they are different applications. Also, we'd have to create a new component (as far as I remember the options) just to change the order of day, month, and year. That's much effort for very little improvement.

>  tasks can be clicked, but clicking them doesn't do anything

Now the cursor is no pointer anymore when hovering the tasks.

>  When starting a job, status doesn't update itself on completion (but it does upon refresh, site reload, abortion etc.)

Now the task list will be updated every minute.

>  input-fields need to check whether their input is valid.

This is definitively on our TODO list.

pkg/importer/scheduler.go
```diff
@@ -0,0 +1,425 @@
...
+		tasks[i] = t
+		i++
+	}
+	s.runningTaskLock.Unlock()
+	for _, t := range tasks {
+		if err := s.AbortTask(t); err != nil {
+			slog.Error("could not abort task", "err", err)
+		}
+	}
+}
```
koplas, line 425: Replace all locks with waiting for all goroutines to finish. There is a race condition while closing the database connection while the download handler still uses it.

pkg/web/sources.go
```diff
@@ -0,0 +1,659 @@
...
+	}
+
+	if endRange, ok := ctx.GetPostForm("start_range"); ok {
+		endRangeTime, err := time.Parse("2006-01-02", endRange)
+		if err != nil {
+			ctx.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
+		} else {
+			jobConfig.EndRange = &endRangeTime
+		}
+	}
```
koplas, line 332: Duplicate code in `addJob`. Change the behavior of `updateJob` or refactor it into a separate function.

pkg/web/sources.go
```diff
@@ -0,0 +1,659 @@
...
+
+	if worker, ok := ctx.GetPostForm("worker"); ok {
+		workerInt, err := strconv.ParseInt(worker, 10, 64)
+		if err != nil {
+			ctx.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
+		} else {
+			jobConfig.Worker = int(workerInt)
+		}
+	}
```
koplas, line 149: Missing validation. This will only result in errors while running a task and not before configuration.

.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "csaf_distribution"]
+	path = csaf_distribution
+	url = git@github.com:csaf-poc/csaf_distribution.git
```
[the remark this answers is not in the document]
koplas: This will be removed, when a new `csaf_distribution` version is released.

koplas:
This implementation no longer fits the requirements.

PR #186, opened 2024-06-21
Fix id search

JanHoefelmeyer:


cintek (approved):
Thanks! Works as expected. I just squashed the commits to prevent a commit that is only removing some output.

PR #189, opened 2024-06-24
Improvements for home

cintek:
This PR contains the following improvements:

* Show headers and footers of cards always on same height for better overview
* Add number of comments and SSVC to show that people already worked on an advisory
* Added hint in lower right corner of "New documents" cards to show if a doc is the first of a completely new advisory or not
* Use different formatting of comments in the "Recent activities" cards so its easier to distinguish them from document titles
* Let users click on cards in "New document" to go to the advisory view

### DIFF
--- a/client/src/lib/Home/CustomCard.svelte
+++ b/client/src/lib/Home/CustomCard.svelte
@@ -9,19 +9,27 @@
 -->
 
 <script lang="ts">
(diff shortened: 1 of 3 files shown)


JanHoefelmeyer (approved):
LGTM & works

PR #191, opened 2024-06-25
Generate TLP expression directly from publisher/TLP infos from the tokens.

s-l-teichmann:
Before the publisher / TLP information from the auth token where transformed into the postfix expression language and then parse into a parse tree to be used to generate the query SQL.

With this PR the publisher / TLP information from the auth token is directly transformed into the parse tree w/o the parsing.


JanHoefelmeyer (approved):
LGTM & works

PR #192, opened 2024-06-26
Show only a single (the most important) CVSS value

cintek:
To not overwhelm the user with information on the Home page we just show a single CVSS value in the area with new advisories which is the one from the `critical` column.


JanHoefelmeyer (approved):
LGTM & works

PR #194, opened 2024-06-27
Support for full text search on comments

s-l-teichmann:
This PR adds support for full text search on the comments of documents.

There are two ways to search the comments:

1 - `csearch`: Like **search** on documents with second language argument for *web like search*
2 -  `mentioned`: Short version with only on argument. more for exact phrase search.

Use  e.g. `"Das ist der Suchtext" german csearch` as a predicate in queries.

In document mode all comments of a document are searched
and in advisory mode all comments of all documents of the advisory.

This PR also introduces the `me` keyword. It will be replaced by the current user's name.


JanHoefelmeyer (approved):
LGTM & works

PR #199, opened 2024-07-01
Use strings from known advisory to minimize dedup efforts

s-l-teichmann:
We de-duplicate strings in CSAF documents as they are highly redundant.
Therefore we have a table of unique strings.

This PR makes use of the  fact that if we already have previous revisions of a advisory over 90% of the strings used in the new one are already known. So we don't have to look up all strings individually for duplicates only the one that are new. The IDs of the already known strings can be re-used.


JanHoefelmeyer (approved):
LGTM & works

Issue #121, opened 2024-05-06
Diff: Let users upload a document to compare it with another one


cintek:
Implement that users can compare one  document which they have on their computers with a document from the database.  And to compare two documents they have on their computer to each other.

This needs changes in frontend and backend as the backend is creating the diff.

s-l-teichmann:
Ideas for the backend:
- A user can upload a document to user-specific pool of documents.
- The documents in that only live a certain amount of time (2h e.g).
- At end of that time period these temp documents are removed from the pool
- Per user there should be a configurable maximal total file size (2GB e.g as default)
- For the whole pool there should also be a configurable size limit (10GB e.g as default)
- The upload documents should only be accessible for the uploader
- The uploaded documents should be referenced by a unique identifier (UUID)
- The /api/diff endpoint should recognize these to load the respective documents from the pool.

cintek:
Feature added with #202

PR #202, opened 2024-07-02
Allow users to compare different advisories

### LINKED ISSUE #120: Diff: Compare documents from different advisories
Implement a feature so that users can compare random documents.

cintek:
Resolves #120 and #121 

TODO:
- [x] Add a first version of a feature that allows users to select different documents and compare them
- [x] Prepare backend so users can upload a document for a diff
- [x] Let users upload a document for diff in the client
- [x] Remove Git diff components
- [x] Remove `Compare` from sidebar
- [x] Refactor, e.g. rename diff components as there is no need to differ between components for the git and the JSON diff


cintek:
@s-l-teichmann I added you to the assignes 'cause the backend shall offer the temporary upload of documents for a diff.

s-l-teichmann:
@cintek PR #207 provides the needed backend infrastructure.

cintek:
@JanHoefelmeyer can you please test the following things:

- Compare different versions of the same advisory (was implemented before this PR but I had to change a few things there so another test is necessary)
- Compare different advisories/documents from `Search`
- Upload documents (no import) and compare them with each other and with other advisories/documents

JanHoefelmeyer (approved):
Functionally works. UI may need further refinements, but within later iterations/after more discussion

cintek:
Thanks for testing! What I forgot in the TODO list is testing the configuration (settings expiration limit etc.). Can you please also test that?

PR #204, opened 2024-07-04
Add events endpoint to allow filtering and selection similiar to the documents endpoint

s-l-teichmann:
This PR implements  filtering and selecting on events like the /api/documents endpoint does.

Supports `query`, `order`, `limit`, `offset` and `count`.

Example:

Returns events which are not an import and are done in the last hour.

```
curl -s -D /dev/stderr \
   --get \
   --data-urlencode 'limit=10' \
   --data-urlencode 'count=true' \
  --data-urlencode 'query=$event import_document events != now 1h duration - $time <=' \
   http://localhost:8081/api/events | jq .
```

### DIFF
--- a/pkg/database/query/parser.go
+++ b/pkg/database/query/parser.go
@@ -449,9 +449,11 @@ func (st *stack) pop() *Expr {
 	panic(parseError("stack empty"))
 }
 
-func (st stack) top() *Expr {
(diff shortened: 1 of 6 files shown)


JanHoefelmeyer (approved):
LGTM & works

PR #207, opened 2024-07-04
Temp storage for documents and enble diff to use these temp docs.

s-l-teichmann:
This PR consists of two parts:

1. Each user is allowed to have some temporary documents on the server.
2. The diff endpoint is enabled to use this temp docs together with the ones from the database.

The temp docs are managed with the `/api/tempdocuments` endpoint.


JanHoefelmeyer (approved):
LGTM & works

PR #208, opened 2024-07-07
Replace fulltext with trigram search

s-l-teichmann:
This PR replaces the full text search with a trigram one.

Before this we were not able to find certain strings in the database.
Be had matches for e.g. 'log' and 'log4j' but not for 'log'.


ThomasJunk (approved):
LGTM

PR #214, opened 2024-07-09
Simplify parameter parsing in endpoints a bit.

s-l-teichmann:
This PR reduces the boiler plate code like
```
idS := ctx.Param("id")
id, err := strconv.ParseInt(idS, 10, 64)
if err != nil {
    ctx.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
    return
}
``` 
often found in the endpoints to
```
id, ok := parse(ctx, toInt64, ctx.Param("id"))
if !ok {
    return
}
```
It's not an optimal solution in any situation (we might have a look at Gin's bind feature in some cases),
but a little improvement concerning code duplication.

### DIFF
--- a/pkg/web/comments.go
+++ b/pkg/web/comments.go
@@ -13,7 +13,6 @@ import (
 	"errors"
 	"log/slog"
 	"net/http"
-	"strconv"
 	"time"
 
 	"github.com/gin-gonic/gin"
@@ -25,10 +24,8 @@ import (
 )
 
 func (c *Controller) createComment(ctx *gin.Context) {
-	docIDs := ctx.Param("document")
-	docID, err := strconv.ParseInt(docIDs, 10, 64)
-	if err != nil {
-		ctx.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
+	docID, ok := parseInt(ctx, ctx.Param("document"))
+	if !ok {
 		return
 	}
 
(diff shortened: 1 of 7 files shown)


koplas:
I make use of optional parameters inside the source manager; it looks like this:
```Go
if rate, ok := ctx.GetPostForm("rate"); ok {
	rateFloat, err := strconv.ParseFloat(rate, 32)
	if err != nil {
		ctx.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
		return
	} else {
		jobConfig.Rate = &rateFloat
	}
}
```
This returns a bad request only if the parameter exists and cannot be parsed. I would need an additional optional parameter to use this feature.

s-l-teichmann:
> I make use of optional parameters inside the source manager; it looks like this:
> 
> ```go
> if rate, ok := ctx.GetPostForm("rate"); ok {
> 	rateFloat, err := strconv.ParseFloat(rate, 32)
> 	if err != nil {
> 		ctx.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
> 		return
> 	} else {
> 		jobConfig.Rate = &rateFloat
> 	}
> }
> ```
> 
> This returns a bad request only if the parameter exists and cannot be parsed. I would need an additional optional parameter to use this feature.

Don't do `else` if the if block ends with a `return`! With this its something like https://github.com/ISDuBA/ISDuBA/blob/de0221b70bcb81025e8f9962bd2c9b08edd7f85b/pkg/web/documents.go#L250

JanHoefelmeyer (approved):
LGTM & works for me

PR #226, opened 2024-07-15
Add another missing index to document_txts

s-l-teichmann:
During the performance analysis of the COUNT statement it was discovered that another join index was missing for good use of the trigram index during text search.


JanHoefelmeyer (approved):
LGTM & works

ThomasJunk (commented):
LGTM

PR #227, opened 2024-07-15
Allow keycloak version to be manually set, update default to 25.0

### LINKED ISSUE #166: keycloak, move to current 24.x and consider moving to keycloak 25
`docs/scripts/installkeycloak.sh`  has `keycloak-24.0.1` hardcoded.

There is https://www.keycloak.org/archive/downloads-24.0.5.html so the version number in the script should be in a variable.

We should also consider moving to
* https://www.keycloak.org/2024/06/keycloak-2500-released.html

koplas:
Closes #166


JanHoefelmeyer:
Doesn't work yet due to some changes made by keycloak. I'll fix that quickly, but someone else should review this (since all changes were made by me)

JanHoefelmeyer:
Should work now, seems like only the health-endpoints were changed.

Some background info, in order to check whether Keycloak is online, the scripts curl their health checkpoints, see https://www.keycloak.org/server/health.

Thus far, those were reachable via the Keycloak port, but since Keycloak 25, those were moved to port 9000 (by default, configurable). 

So now configurekeycloak.sh will check on port 9000 whether Keycloak is running unless -l/--live flags specify a different port.

When testing, it would be best if the default as well as a different version using a different health port was used.

PR #228, opened 2024-07-15
Update Java Version to 21

### LINKED ISSUE #169: Dev Setup Java 21 from Ubuntu?
In one of my test setups java 21 from Ubuntu 24 LTS (aka apt-install default-jre ) works fine as far as I can see. We should consider switching the installutilities.sh to it.  (why it is in the "utilities" install file anyway)

koplas:
This will be required for Keycloak 25.
Closes #169


JanHoefelmeyer (approved):
LGTM & works

PR #231, opened 2024-07-16
Add health endpoint

koplas:
This can be useful for the client to detect if the connection to the server is lost.


koplas:
Currently, the client can't differentiate between an internal server error caused by a database error or a malformed request, and the server is unreachable. All other endpoints require authentication, which doesn't allow checking the server status to check if the authentication failed.

This is useful for detecting the error's reason for a failed login or a timeout.

JanHoefelmeyer (approved):
LGTM & works

s-l-teichmann:
@JanHoefelmeyer  Argh!! Please revert the merge. We need to discuss this more deeply!

Issue #232, opened 2024-07-16
`$title foobar ilike` advanced search not working


s-l-teichmann:
The `ilike` search does not work atm. I'll have a look at this.

PR #233, opened 2024-07-16
Let ilike commands behave like mentioned and search

s-l-teichmann:
Solves #232

This fixes an inconsistency between `ilike`/`ilikepid` and `search`/`mentioned`.

When the full text search was replaced by the trigram search the search terms were replaced by
`ILIKE` statements on the SQL side. `ILIKE` statements need anchoring with `%`  to match at any
position. To spare the users from wrapping their search terms in '%' this is done by the engine.
On the other side the `ilike`/`ilikepid` were already `ILIKE` base and required explicit `%` bracketing.

So the correct search term would actually be `$title %foobar% ilike`. To fix this inconsistency this PR
removes the need to issue these `%`s.

### DIFF
--- a/pkg/database/query/sqlbuilder.go
+++ b/pkg/database/query/sqlbuilder.go
@@ -207,12 +207,17 @@ func (sb *SQLBuilder) nowWhere(_ *Expr, b *strings.Builder) {
 	b.WriteString("current_timestamp")
 }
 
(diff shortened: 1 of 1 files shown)


JanHoefelmeyer (approved):
LGTM & works

PR #236, opened 2024-07-18
Don't allow reviewer comment on state new or read

### LINKED ISSUE #206: Reviewer can attempt to comment while advisory is new/read, but it breaks
Error message: "Could not create comment. The request sent could not be understood."

This also expands the comment-box to overlap with the history/events.

koplas:
Closes #206


JanHoefelmeyer (approved):
LGTM & works

PR #237, opened 2024-07-18
Change advisory state to new, if a new document was inserted

### LINKED ISSUE #143: Import via bulkimporter does not set advisory state to new
When importing an advisory that is set to delete, the workflow status will not be changed, however, the event-view will mention the status being changed.

Assuming that the workflow status should also be set to new, since new information is available.

koplas:
Closes #143

### DIFF
--- a/pkg/database/migrations/012-insert_as_new.sql
+++ b/pkg/database/migrations/012-insert_as_new.sql
@@ -1,0 +1,41 @@
+---- This file is Free Software under the Apache-2.0 License
+-- without warranty, see README.md and LICENSES/Apache-2.0.txt for details.
+--
+-- SPDX-License-Identifier: Apache-2.0
+--
+-- SPDX-FileCopyrightText: 2024 German Federal Office for Information Security (BSI) <https://www.bsi.bund.de>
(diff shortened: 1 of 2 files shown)


pkg/database/migrations/012-insert_as_new.sql
```diff
@@ -0,0 +1,43 @@
...
+
+-- create_advisory checks if the new document is newer than the old one.
+CREATE OR REPLACE FUNCTION create_advisory() RETURNS trigger AS $$
+    DECLARE
+        old_id           int;
+        old_rev_length   int;
+        old_release_date timestamptz;
+    BEGIN
+        -- Ensure having an advisories record.
+        INSERT INTO advisories (tracking_id, publisher)
```
s-l-teichmann, line 17: Instead of the `UPDATE` below we might use this `INSERT` with an `ON CONFLICT` clause that updates the state field to `'new'`

s-l-teichmann (changes requested):

s-l-teichmann (approved):

JanHoefelmeyer (approved):
LGTM & works

PR #241, opened 2024-07-19
Product vulnerabilities

ThomasJunk:
Re-introduces the product vulnerability crosstable.

New:

There are now two modes:

- Display only the CVEs which are _relevant for display_ recurring on the implementation of the `four_cves` column 
- Full display of all products ⨯ vulnerabilities

Legend is now on top  instead of right.

This is a (small) improvement to the status quo.

### DIFF
--- a/client/src/lib/Advisories/Advisory.svelte
+++ b/client/src/lib/Advisories/Advisory.svelte
@@ -35,6 +35,7 @@
   let loadEventsError = "";
   let loadAdvisoryVersionsError = "";
   let loadDocumentError = "";
+  let loadFourCVEsError = "";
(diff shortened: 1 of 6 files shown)


JanHoefelmeyer (changes requested):
Jumping to a point via the product overview is persistent on multiple visits to an advisory.

s-l-teichmann:
> Jumping to a point via the product overview is persistent on multiple visits to an advisory.

@JanHoefelmeyer As @ThomasJunk  is currently not available atm may you have a look at this yourself?!

JanHoefelmeyer (approved):
LGTM & works now

PR #253, opened 2024-07-24
Source Manager Mockup

koplas:
This draft explores the requirements of the source manager.


koplas:
A new UI will be required for the source manager.

PR #266, opened 2024-07-29
Only load events if allowed

### LINKED ISSUE #240: Error message on front page for roles that are not allowed to view recent events
This affects Admins and Importer.

koplas:
Closes #240

### DIFF
--- a/client/src/lib/Home/RecentActivities.svelte
+++ b/client/src/lib/Home/RecentActivities.svelte
@@ -257,7 +257,7 @@
   });
 </script>
 
-{#if $appStore.app.isUserLoggedIn}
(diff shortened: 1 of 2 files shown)


JanHoefelmeyer (approved):
This works, but other roles likely also want to have a recent activity feed if they are involved with advisories. (E.g. pinging admins for info)

So instead of just disabling the view, we might want to give access to events instead.

JanHoefelmeyer:
Admin is not the only one affected, Importer as well.

Also now pinging admins will alert them on the homepage, but it will lead to an error (not allowed to see comments.)

Lastly, they'd need to be able to see comments for this to be worth it, which has some heavier implications.

koplas:
> Admin is not the only one affected, Importer as well.

The importer role is intended for uploads only.

JanHoefelmeyer (approved):
LGTM & works

PR #267, opened 2024-07-29
Fix edit comment permissions

### LINKED ISSUE #93: When are comments allowed to be changed?
Currently, a user can edit comments even when they no longer have permission to write a new one. We need to contemplate whether this is the intended way to handle comments, and if not, what would be the intended way to handle it

koplas:
Only allow to edit comments, if it is allowed to create a new comment. Closes #93


JanHoefelmeyer (approved):
You can still edit comments without changing workflowstates if you couldn't write new ones without changing workflowstates, but I think this is not really an issue.

PR #271, opened 2024-07-29
Show details for ErrorMessage

### LINKED ISSUE #138: Error messages should give available technical details to aid administrators for analysis
If there are technical information available, in an error situation, they should be presented to the user as additional detail that they can use to give to their technical contacts.

This often includes:
 * datetime where the problem occurred
 * the status code of a backend answer, a translation what it means
 * the body of a backend answer
 * the url or parts of the request

koplas:
Closes #138


cintek (changes requested):
An important thing I forgot is that Details should not look like a link. It would be better to use a chevron icon to indicate that you can open and collapse the details.

A nice-to-have: Now, if you open the details it happens the "x" is moving. It would be better if it would stay at one place. Maybe you could place it in the upper left corner.

cintek (approved):

PR #272, opened 2024-07-29
Better sorting of null CVSS values

### LINKED ISSUE #163: Sort documents without any CVSS value better
Since #154 we can sort new documents by criticality better if they have at least one CVSS value.  This will be used to show the most interesting new documents. 

Currently having no CVSS value will sort the documents to be last. Can we sort it better?

This is a question of how interesting new documents are with out CVSS value to the users that want to evaluate them?

Are they most interesting and should be looked upong first. because they could of very high criticalilty? Or will they usually be of medium interest, like to be sorted between medium <7 and  high?

koplas:
CVSS values that are null are now sorted like they have the score of 5.

Closes #163

### DIFF
--- a/pkg/database/query/sqlbuilder.go
+++ b/pkg/database/query/sqlbuilder.go
@@ -372,7 +372,7 @@ func (sb *SQLBuilder) CreateOrder(fields []string) (string, error) {
 		case "cvss_v2_score", "cvss_v3_score", "critical":
 			b.WriteString("COALESCE(")
 			b.WriteString(field)
-			b.WriteString(",0)")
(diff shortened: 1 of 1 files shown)


bernhardreiter:
What is your reason to select 5? (So far I do not remember that we have talked or decided about a specific value yet.)

koplas:
> What is your reason to select 5? (So far I do not remember that we have talked or decided about a specific value yet.)

There is no specific reason. This pull request is for discussing the sorting of CVSS values.

s-l-teichmann:
Assuming that a not set CVSS value equals 5 is plain wrong.

PR #279, opened 2024-08-01
Use proper links for products and cves

### LINKED ISSUE #242: Advisory overview: Browsers "back" button should allow to scroll to the last point scrolled away from
If A links to B but B not to A, finding a way to get back to A from B would be an improvement.

koplas:
Todo: Save the scroll position before clicking on the link.

Closes #242


ThomasJunk (approved):
LGTM

Issue #83, opened 2024-04-11
Show the user that opening the profile page opens a new tab


JanHoefelmeyer:

bernhardreiter:
A new tab should not be opened by us actively.

cintek:
@bernhardreiter what do you mean by "actively"? At this time the user has to click on the `Profile` tab to open the profile page.

Since the profile page is part of Keycloak IMHO it makes sense to open it in a new tab. And to indicate that a click on the `Profile` tab leads to a new tab our idea was to use an icon (a rectangle with an arrow showing to the top right as in many other applications).

bernhardreiter:
An icon that qualifies the menu entry as one that will lead to a completely different page is an improvement. (As opposed to menu entries that lead to a new page within our main application.)
Login and profile are the same I guess, so both would need that icon.

The question is why we should open a new tab with one of them. Login does not open a new tab IIRC.

bernhardreiter:
Can we give keycloak a backlink, so it will redirect to our application after the editing of the user account is done?

cintek:
> Login and profile are the same I guess, so both would need that icon.

Good argument.

> The question is why we should open a new tab with one of them. Login does not open a new tab IIRC.

Because the login process is a short one but when you navigate through the profile page you could have to go back in the browser many times until you get back to our application. But I guess that is the reason for the next comment:


> Can we give keycloak a backlink, so it will redirect to our application after the editing of the user account is done?

I don't know but in Keycloak you can edit `Personal info`, `Accounty security` and/or `Application` so it wouldn't be good to redirect the user after editing the first one because it could be that the user wants to edit the sedond also.

Another idea: Use an iframe for the profile page. That way the user has still the context of our application (the sidebar) and can navigate to another site very quickly.

bernhardreiter:
Hmm the logout also goes to the keycloak somehow (at least I hope that I get logged out of keycloak). Maybe login, logout and option all need to go under one subpage or sub menu and there we make sure that this is outside.

Add the icon for the time being to profile as a first improvement.

I could imagine that keycloak has the ability to go back to the application somehow.

JanHoefelmeyer:
Reworked the profile page, issue obsolete

JanHoefelmeyer:
Seems like this is no longer obsolete:
https://github.com/ISDuBA/ISDuBA/issues/111#issuecomment-2082173196

ThomasJunk:
The referring issue is closed so I close this one

[...]

PR #280, opened 2024-08-01
doc: Move example environment file to docs, explain values

JanHoefelmeyer:
closes https://github.com/ISDuBA/ISDuBA/issues/83


docs/env.example
```diff
@@ -0,0 +1,33 @@
...
+# SPDX-FileCopyrightText: 2024 German Federal Office for Information Security (BSI) <https://www.bsi.bund.de>
+#  Software-Engineering: 2024 Intevation GmbH <https://intevation.de>
+
+# Where to reach the application's keycloak-instance
+PUBLIC_KEYCLOAK_URL="http://localhost:8080"
+
+# The name of the keycloak-realm managing all ISDuBA accounts
+PUBLIC_KEYCLOAK_REALM="isduba"
+
+# The name of the client that can request the authentification of a user
```
cintek, line 15: Since we have code in a folder named `client` and the application itself is a client I would specify here that a keycloak-client is meant (like it is already in the name of the variable).

cintek (changes requested):

docs/env.example
JanHoefelmeyer: I specified keycloak-client, does this suffice?

cintek (approved):
LGTM

koplas:
@JanHoefelmeyer I would like to close this pull request in favor of #287.

JanHoefelmeyer:
Makes sense

PR #287, opened 2024-08-01
Move client configuration to backend

### LINKED ISSUE #281: Combine client and server config
We should combine the clients .env file and the servers config toml-file into one config file and let the client get its config from the server.

koplas:
Closes #281


pkg/config/config.go
```diff
@@ -144,6 +153,16 @@ type Sources struct {
 	DefaultMessage string `toml:"default_message"`
 }
 
+// Client are the config options for the client.
+type Client struct {
+	KeycloakUrl      string `toml:"keycloak_url"`
+	KeycloakRealm    string `toml:"keycloak_realm"`
```
s-l-teichmann, line 159: If you add json struct tags you can simply send the structure as JSON in the endpoint.

s-l-teichmann (changes requested):

pkg/config/config.go
koplas: Done

pkg/config/config.go
```diff
@@ -313,6 +341,12 @@ func (cfg *Config) fillFromEnv() error {
 		envStore{"ISDUBA_TEMP_STORAGE_FILES_USER", storeInt(&cfg.TempStore.FilesUser)},
 		envStore{"ISDUBA_TEMP_STORAGE_DURATION", storeDuration(&cfg.TempStore.StorageDuration)},
 		envStore{"ISDUBA_SOURCES_DEFAULT_MESSAGE", storeString(&cfg.Sources.DefaultMessage)},
+		envStore{"ISDUBA_CLIENT_KEYCLOAK_URL", storeString(&cfg.Client.KeycloakURL)},
+		envStore{"ISDUBA_CLIENT_KEYCLOAK_REALM", storeString(&cfg.Client.KeycloakRealm)},
+		envStore{"ISDUBA_CLIENT_KEYCLOAK_CLIENTID", storeString(&cfg.Client.KeycloakClientID)},
```
s-l-teichmann, line 346: s/ISDUBA_CLIENT_KEYCLOAK_CLIENTID/ISDUBA_CLIENT_KEYCLOAK_CLIENT_ID/

pkg/config/config.go
```diff
@@ -144,6 +153,16 @@ type Sources struct {
 	DefaultMessage string `toml:"default_message"`
 }
 
+// Client are the config options for the client.
+type Client struct {
+	KeycloakURL      string `toml:"keycloak_url" json:"keycloak_url"`
+	KeycloakRealm    string `toml:"keycloak_realm" json:"keycloak_realm"`
+	KeycloakClientID string `toml:"keycloak_client_id" json:"keycloak_client_id"`
+	UpdateInterval   int    `toml:"update_interval" json:"update_interval"`
```
s-l-teichmann, line 161: Okay, it's for the client,  but it feels a bit off as we use time.Duration for time intervals used by the server.

s-l-teichmann (changes requested):

pkg/config/config.go
koplas: Done

PR #301, opened 2024-08-09
Allow to drag and drop in query designer

koplas:


cintek (changes requested):
The three dots look more like a menu. Maybe we could use this icon which is used often at application where you can sort something (hard to see when you use the dark mode of Github):
grid-dots-vertical-rounded
(this is a custom icon which combines the dot-vertical and the grid-vertical icons)

Another aspect is that we should show the icon only when a user hovers over a row but as the PR is still a draft it may be that you are aware of it.

koplas:
> The three dots look more like a menu. Maybe we could use this icon, which is used often at application where you can sort something (hard to see when you use the dark mode of Github): grid-dots-vertical-rounded (this is a custom icon which combines the dot-vertical and the grid-vertical icons)
> 
> Another aspect is that we should show the icon only when a user hovers over a row, but as the PR is still a draft, it may be that you are aware of it.

Can you include the icon and check its license?
I removed the icon hover because it decreases accessibility (requires a mouse and can confuse the user).

JanHoefelmeyer:
I'd honestly expect Drag&Drop to work via the names, not the icons, which isn't very intuitive. The icons being the same for every single column also seems redundant.  (In addition, as cintek mentioned, the icons indicate a menue).

cintek:
> I'd honestly expect Drag&Drop to work via the names, not the icons, which isn't very intuitive.

I'd expect it to work with both but it is necessary to show the icons (or at least one icon in the row the user is hovering) to indicate that the user is able to drag&drop the items. Otherwise, the user would find this out only by accident but it is important that the user knows how to re-order the items.

cintek:
> Can you include the icon and check its license?

Done with 784df9e8769153ffe00c8bef798e9747cfcd7483.

> I removed the icon hover because it decreases accessibility (requires a mouse and can confuse the user).

That's a good argument. The commit above also uses a gray color for the icon so the icons are a bit less outstanding.

Something else: If you change the order in the query designer and click on "Save" the UI elements of the items are re-ordered again. Although the correct order is visible when you open the query the next time this behavior is irritating because you think that your order wasn't applied.

https://github.com/user-attachments/assets/92678d63-f64c-441e-89fb-fcc093b7b180

cintek:
> Something else: If you change the order in the query designer and click on "Save" the UI elements of the items are re-ordered again. Although the correct order is visible when you open the query the next time this behavior is irritating because you think that your order wasn't applied.

We will put this into an issue.

cintek (approved):
LGTM & works

PR #303, opened 2024-08-10
Construct client config from server config if not explictly set.

s-l-teichmann:
We already have partsof the config needed by the client as config for the server.

This PR sets `client.keycloak_url` to `keycloak.url` if not specified.


koplas:
I removed the `application_uri` in main as this configuration option is not necessary and resulted in multiple confusions in the past. The keycloak URL part looks good to me.

koplas (approved):

PR #308, opened 2024-08-12
Allow to specify multiple order levels

### LINKED ISSUE #302: Query Designer: Add the possibilty to add more than one order criterion
The UI currently offers only the possibility to give one order criterion. This is to coarse:
Assume sorting by `critical` resulting in a list of entries which all have same value.
In this case you may want the more recently updated advisories/documents appear in the front of that list.

The backend already supports this.

koplas:
Closes #302

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -25,6 +25,7 @@
         "@sveltejs/kit": "^2.5.18",
         "@sveltejs/vite-plugin-svelte": "^3.0.0",
         "@types/eslint": "8.56.10",
+        "@types/qs": "^6.9.15",
         "@types/sortablejs": "^1.15.8",
         "@typescript-eslint/eslint-plugin": "^7.17.0",
         "@typescript-eslint/parser": "^7.13.0",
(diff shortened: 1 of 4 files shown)


cintek:
The UI for selecting the order looks a bit unordinary. When using an indicator (like [Vuetify does](https://vuetifyjs.com/en/components/data-tables/data-and-display/#multi-sort)) for the order of the order we would only need one column.

s-l-teichmann:
> The UI for selecting the order looks a bit unordinary. When using an indicator (like [Vuetify does](https://vuetifyjs.com/en/components/data-tables/data-and-display/#multi-sort)) for the order of the order we would only need one column.

This not true ... Think the dashbord where we have only 6 cards. Selecting only the first 6 items with on column order criterion will bring up other results has having more than one right from the start.

cintek:
> This not true ... Think the dashbord where we have only 6 cards. Selecting only the first 6 items with one column order criterion will bring up other results has having more than one right from the start.

Sorry, I didn't make my point clear enough. I meant the columns in the UI used to display the order:
order

cintek (approved):
LGTM & works

PR #309, opened 2024-08-12
Allow drag and drop in query overview

koplas:


pkg/web/queries.go
```diff
@@ -157,6 +157,55 @@ func (c *Controller) createStoredQuery(ctx *gin.Context) {
 	})
 }
 
+func (c *Controller) updateOrder(ctx *gin.Context) {
+	updateSQL := `UPDATE stored_queries SET num = ( CASE`
+	err := ctx.Request.ParseMultipartForm(1000 * 1000)
```
s-l-teichmann, line 162: Maybe it is a better idea to pass a JSON document to this endpoint.

pkg/web/queries.go
```diff
@@ -157,6 +157,55 @@ func (c *Controller) createStoredQuery(ctx *gin.Context) {
 	})
 }
 
+func (c *Controller) updateOrder(ctx *gin.Context) {
+	updateSQL := `UPDATE stored_queries SET num = ( CASE`
```
s-l-teichmann, line 161: This generated case feels a bit tricky. I would prefer a transaction running a batch update in it.

s-l-teichmann (changes requested):

pkg/web/queries.go
```diff
@@ -157,6 +157,48 @@ func (c *Controller) createStoredQuery(ctx *gin.Context) {
 	})
 }
 
+func (c *Controller) updateOrder(ctx *gin.Context) {
```
s-l-teichmann, line 160: Almost ... (untested)
```Go
func (c *Controller) updateOrder(ctx *gin.Context) {
	type queryOrder struct {
		ID    int64 `json:"id"`
		Order int64 `json:"order"`
	}
	var orders []queryOrder
	if err := ctx.ShouldBindJSON(&orders); err != nil {
		ctx.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
		return
	}

	const (
		prefix   = `UPDATE stored_queries SET num = $2 WHERE id = $1 AND `
		adminSQL = prefix + `(definer = $3 OR global)`
		userSQL  = prefix + `definer = $3`
	)
	var updateSQL string
	if c.hasAnyRole(ctx, models.Admin) {
		updateSQL = adminSQL
	} else {
		updateSQL = userSQL
	}

	batch := &pgx.Batch{}
	definer := ctx.GetString("uid")
	for _, order := range orders {
		batch.Queue(updateSQL, order.ID, order.Order, definer)
	}

	if err := c.db.Run(
		ctx.Request.Context(),
		func(rctx context.Context, conn *pgxpool.Conn) error {
			tx, err := conn.Begin(rctx)
			if err != nil {
				return err
			}
			defer tx.Rollback(rctx)
			if err := tx.SendBatch(rctx, batch).Close(); err != nil {
				return err
			}
			return tx.Commit(rctx)
		}, 0,
	); err != nil {
		slog.Error("database error", "err", err)
		ctx.JSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
	}
}

```

s-l-teichmann (changes requested):

cintek (approved):
LGTM & works

PR #311, opened 2024-08-13
Add initial source manager ui

koplas:
This initial version is not pretty and looks more like a debug menu. It provides a basis for further UI discussions.


cintek (approved):
LGTM & works

PR #313, opened 2024-08-13
Stored queries types

s-l-teichmann:
This PR replaces the boolean `advisories` field of stored queries with a field `kind` with the possible values `advisories`, `documents` and `events`.

@cintek The client needs the corresponding adjustments


JanHoefelmeyer (approved):
works

Issue #162, opened 2024-06-12
Adjust view for users without group


JanHoefelmeyer:
Users without group are treated as if they're in a group with priviledges {"*": ["WHITE"]}.

Users with these explicit groups will be shown this, users who obtain them by default will not see anything. This is inconsistent and a default value should also be shown in the profile

JanHoefelmeyer:
Since the default group behaviour is defined on the server, this could probably be best done after https://github.com/ISDuBA/ISDuBA/issues/281 is implemented.

PR #316, opened 2024-08-14
Fix restrictions

JanHoefelmeyer:
Fixes two bugs that allowed users without groups to bypass permission checks.

Also adds an unit test that checks whether permissions are properly denied when they should be.

Also closes https://github.com/ISDuBA/ISDuBA/issues/162


s-l-teichmann:
LGTM & works

PR #317, opened 2024-08-14
Source Manager: Add custom header values to HTTP requests

s-l-teichmann:
This PR adds support for custom http header values per source to the source manager downloads.


ThomasJunk (approved):
LGTM

JanHoefelmeyer (approved):
LGTM & doesn't add bugs to existing structures as far as I can tell

PR #319, opened 2024-08-16
Source manager: More configuation options

s-l-teichmann:
This PR add support for per source flags of `insecure`, `age`, `strict_mode` and `ignore_patterns`.


ThomasJunk:
lgtm. Testwise persisted data. Seems to work.

JanHoefelmeyer (approved):
Doesn't inhibit any existing structures as far as I could tell

PR #328, opened 2024-08-19
Source manager: Add support for client certificates

s-l-teichmann:
Add client certificate support.
Fields public key, private key and passphrase.
The passphrase and the private key are stored encrypted.


ThomasJunk:
LGTM

docs/isdubad-config.md
```diff
@@ -109,6 +109,14 @@ Valid values for `tlps` are the [Traffic Light Protocol](https://en.wikipedia.or
 - `feed_importer`: Name of the user that is doing the feed imports. Defaults to `feedimporter`.
 - `publishers_tlps`: Rules what the feed import is allowed to import. Defaults to `{ "*" = [ "WHITE", "GREEN", "AMBER", "RED" ] }`
 - `default_message`: The message that should be displayed inside the source manager.
+- `aes_key`: A crypto token in form of 64 character long hex string used to encrypt security sensitive\
+   fields in the database like passphrasses and private keys. Defaults to "".\
+   If an empty token is given, the server will generate a new one every time it starts.\
```
bernhardreiter, line 114: Leaving this field empty generates a new token each time, which means that the saved passphrases will not be accessible anymore and lost. What about stopping the server if the field is empty instead to prevent a later error situation which is hard to understand?

bernhardreiter:
For security reasons, the backend should not give out the private key materials or passwords anymore.   So sending this to the backend is okay, where is it encrypted and saved, but reading should not be possible. I have not checked the surrounding code deep enough yet.

The attack to defend against is:
If somehow cracks Peter's account info or get access to his session, they could read out the private key material or passwords for other authentification.

bernhardreiter:
Is saving the `aes_key` helping much for development reasons? I am asking myself if it would be a simplification (of code and setup variants) if we'd only allow for an external file in all cases.

s-l-teichmann:
> Is saving the `aes_key` helping much for development reasons? I am asking myself if it would be a simplification (of code and setup variants) if we'd only allow for an external file in all cases.

I don't understand this question.

docs/isdubad-config.md
s-l-teichmann: > Leaving this field empty generates a new token each time, which means that the saved passphrases will not be accessible anymore and lost. What about stopping the server if the field is empty instead to prevent a later error situation which is hard to understand?

I have considered this and I am still not sure if this good idea or not as not all potential users of ISDuBA are going to use the secured stuff.

bernhardreiter:
> > Is saving the `aes_key` helping much for development reasons? I am asking myself if it would be a simplification (of code and setup variants) if we'd only allow for an external file in all cases.
> 
> I don't understand this question.

Sorry, I somehow send this off without mentioning that it is about the variant that saves it in the config file. 

Documenting what you said to me since then:
The variant is offered to be more symmetric with other sensitive field which can also be saved in the toml file. Potentially they could also gain the `@` possibility. `aes_key`  is slightly different and more sensitive than the password for the database.

(It is clarified for me now.)

koplas (approved):
LGTM

PR #329, opened 2024-08-19
Make queries on dashboard configurable

cintek:
Admins will be able to select queries which will be shown on the dashboard of all users belonging to specific roles. Then the users can decide to overwrite those queries with own ones.

Solves #361.


pkg/database/migrations/020-stored_queries_dashboard_roles.sql
```diff
@@ -0,0 +1,23 @@
...
+
+CREATE TYPE stored_queries_roles AS ENUM (
+    'editor', 'reviewer', 'auditor', 'source-manager', 'importer', 'admin'
+);
+
+ALTER TABLE stored_queries
+    ADD COLUMN dashboard bool NOT NULL DEFAULT FALSE,
+    ADD COLUMN role stored_queries_roles;
+
+CREATE TABLE default_query_exclusion (
```
s-l-teichmann, line 17: Needs to be in 000-setup.sql, too

s-l-teichmann (changes requested):

pkg/database/migrations/020-stored_queries_dashboard_roles.sql
cintek: Added with ea1dae4

PR #330, opened 2024-08-19
Extend source ui configuration

koplas:


ThomasJunk:
As a _draft implementation_ of the UI this will be ok.

But there is nice GOTCHA in the implementation of how the server deals with `age`:
- The required `input format` is `string` like `17520h` (which is an acceptable `duration` in go).
- The `output-format` of the source-endpoint is `nanoseconds` like `63072000000000000` which the client has somehow to make sense of like `17520h`.
- 
Which leads to the task of the client doing some guesswork :neutral_face: 

Besides that it might be a good idea to consider to refactor the two forms into one used in `SourceEditor.svelte`. But this might only be a small win. I'm not sure.

ThomasJunk (approved):
LGTM and works (with a grain of salt).

s-l-teichmann:
> As a _draft implementation_ of the UI this will be ok.
> 
> But there is nice GOTCHA in the implementation of how the server deals with `age`:
> 
>     * The required `input format` is `string` like `17520h` (which is an acceptable `duration` in go).
> 
>     * The `output-format` of the source-endpoint is `nanoseconds` like `63072000000000000` which the client has somehow to make sense of like `17520h`.
> 
> 
> Which leads to the task of the client doing some guesswork 😐
> 
> Besides that it might be a good idea to consider to refactor the two forms into one used in `SourceEditor.svelte`. But this might only be a small win. I'm not sure.

PR #337 addresses this.

PR #336, opened 2024-08-22
Source manager: Implement strict mode

s-l-teichmann:
This PR implements the strict mode incl. checking filenames.


ThomasJunk (approved):
LGTM

PR #339, opened 2024-08-22
Source manger: store signature, filename and statistics in database

s-l-teichmann:
This PR stores the signature, the filename and the fail stats into the database.


ThomasJunk (approved):
LGTM & application works

PR #347, opened 2024-08-27
Update eslint dependencies

cintek:
@JanHoefelmeyer can you please test if the advisory table and the diff components are still working with these changes?


JanHoefelmeyer (approved):
Both work

PR #348, opened 2024-08-27
Diff: Evaluate empty TLPs as access denial

s-l-teichmann:
If we need data from the database we should always evaluate empty TLPs as a access denial.


JanHoefelmeyer (approved):
LGTM & works

PR #350, opened 2024-08-28
Source manager: stats support for sources and feeds

s-l-teichmann:
This PR add stats support for feeds and sources.

* There is a new endpoint `GET`  `/api/sources/:id` to fetch only one source.
* `GET` `/api/sources`, `/api/sources/:id`, `/api/sources/:id/feeds` and `/api/sources/feeds/:id` all supports now an optional query parameter `?stats=true` which add stats output to the returned JSON.
* **Important**: This PR also fixes a bug which led to downloading advisories which are already in the database.

### DIFF
--- a/pkg/sources/manager.go
+++ b/pkg/sources/manager.go
@@ -85,6 +85,25 @@ const (
 	SourceDeactivated
 )
 
+// SourceInfo are infos about a source.
+type SourceInfo struct {
+	ID                      int64
+	Name                    string
+	URL                     string
+	Active                  bool
+	Rate                    *float64
+	Slots                   *int
+	Headers                 []string
+	StrictMode              *bool
+	Insecure                *bool
+	SignatureCheck          *bool
+	Age                     *time.Duration
+	IgnorePatterns          []*regexp.Regexp
+	HasClientCertPublic     bool
+	HasClientCertPrivate    bool
+	HasClientCertPassphrase bool
+}
+
 func (sur SourceUpdateResult) String() string {
 	switch sur {
 	case SourceUnchanged:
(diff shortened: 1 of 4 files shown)


pkg/sources/manager.go
```diff
@@ -263,80 +299,136 @@ func (m *Manager) Run(ctx context.Context) {
...
+			fiCh <- nil
 			return
 		}
-		fn(f.label, f.url, f.rolie, config.FeedLogLevel(f.logLevel.Load()))
-		errCh <- nil
+		var st *Stats
+		if stats {
+			st = new(Stats)
+			f.addStats(st)
+		}
```
koplas, line 421: For feeds and sources, it can be useful to extend the function to only return the statistics. This can be beneficial, if we later implement polling of the statistics on the frontend.

koplas (approved):
LGTM (I added some remarks about future extensions of the statistic support).

pkg/sources/manager.go
```diff
@@ -85,6 +85,42 @@ const (
 	SourceDeactivated
 )
 
+// Stats are some statistics about feeds and sources.
+type Stats struct {
```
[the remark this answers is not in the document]
s-l-teichmann: > Will a failing downloads statistic be added later?

Yes, on different way. The source manger does not remember these in-core. These values could be taken from the database any time.

pkg/sources/manager.go
s-l-teichmann: > For feeds and sources, it can be useful to extend the function to only return the statistics. This can be beneficial, if we later implement polling of the statistics on the frontend.

The other values are cheap enough to be polled.

PR #352, opened 2024-08-31
Source manager: Add 'status' field to source reflecting internal misconfigs.

s-l-teichmann:
This PR adds a `status` field to the output of the endpoints `GET` `/api/sources` and `/api/sources/:id` indicating that the  internal configuration is wrong and the source is deactivated
due to this reason. The `status` field is not in JSON object if every thing is fine.

Currently the only reason triggering this is that the client cert config is invalid
(e.g. wrong passphrase to unlock the private key). If we maybe get more reasons
 in the future  we should consider to replace it with a more sophisticated mechanism.


koplas:
For usability reasons, I would prefer a client certificate validation on source creation or update.

pkg/web/sources.go
```diff
@@ -49,6 +49,7 @@ type source struct {
 	Name                 string         `json:"name" form:"name" binding:"required,min=1"`
 	URL                  string         `json:"url" form:"url" binding:"required,min=1"`
 	Active               bool           `json:"active" form:"active"`
+	Status               string         `json:"status,omitempty"`
```
koplas, line 52: We should use an string array to collect multiple errors on future versions.

s-l-teichmann:
> For usability reasons, I would prefer a client certificate validation on source creation or update.

On Update this is already done. The Update returns `deactivated` if the source is configured badly.

On creation this fine as we cannot create active sources. We always have to activate is via an update.
So the remarks about updating holds. We can discuss about that this could be more elaborated.

Keep in mind that this deactivation could also occur at boot time. If we start up with a config that does not work we deactivate it, too. At this point in time there is no UI to send this to the users.

koplas (approved):
Approving this as more user-friendly ways are hard or impossible to implement.

JanHoefelmeyer (approved):
LGTM & works 
(The Error I used was artificially created, but should emulate a real situation)

PR #368, opened 2024-09-05
Add feed log pagination

koplas:


cintek (changes requested):
There is a problem when changing the logs per page:

1. Add Sick as a source and wait for downloads to finish
2. Go to the logs of the feed -> there are 12 pages when 10 logs per page are shown
3. Navigate to the 12th page
4. Select 25 logs per page -> problem: there are only 5 pages but you are still on the 12th page
5. Click multiple times on the button to step one page back -> you can navigate to page 8 but then you can't navigate to page 7 or lower (only if you manually enter the page or click on the button to go to the first page)

It seems that the current is not correctly set when you change the logs per page.

cintek (approved):
Thanks! Works.

PR #369, opened 2024-09-05
Source save improve

koplas:


cintek (dismissed):
If you change a source and revert the change the save button is still active although the final options do not differ from the old options.

cintek (approved):
is good enough for now

PR #373, opened 2024-09-05
Fix the download of directory feeds

koplas:


s-l-teichmann:
Thanks for fixing this! The `changes.csv` is indeed needed for directory based feeds. Must be got lost during development/refactoring. Re-tested it with redhat.com (which I've done before ... shame on me for breaking it). Little nit. I've changed the simple string concat to `url.JoinPath`.

LGTM & works.

PR #375, opened 2024-09-06
Restyle FeedView

luisa-beerboom:
Added minmum widths to selector columns to ensure the selection remains visible. 
Centered active checkboxes in active column to make it look better.


koplas (approved):
LGTM

PR #376, opened 2024-09-06
Only enable save if source is different

### LINKED ISSUE #374: Source management: Enable save button only if there really is a change
If you change a source and revert the change the save button is still active although the final options do not differ from the old options.

koplas:
Closes #374

### DIFF
--- a/client/src/lib/Sources/SourceEditor.svelte
+++ b/client/src/lib/Sources/SourceEditor.svelte
@@ -69,6 +69,8 @@
     ignore_patterns: [""]
   };
 
+  let oldSource = structuredClone(source);
+
   const dtClass: string = "ml-1 mt-1 text-gray-500 md:text-sm dark:text-gray-400";
   const ddClass: string = "break-words font-semibold ml-2 mb-1";
 
@@ -80,6 +82,8 @@
       if (fillAgeDataFromSource) {
         fillAgeDataFromSource(source);
       }
+      await updateSourceForm();
+      oldSource = structuredClone(source);
       sourceEdited = false;
     } else {
       loadSourceError = result.error;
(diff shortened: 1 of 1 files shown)


JanHoefelmeyer:
Source considers itself changed if an empty ignore pattern is added. (I think that's not really an issue though)
Otherwise, LGTM & works

cintek:
@koplas can you please solve the conflicts?

cintek (approved):
LGTM

cintek (changes requested):
Does not work for e.g. `rate` or `insecure`. Probably because the initial value is undefined or null but when you select and deselect a checkbox or enter a character and remove it from an input the new value is `false` or an empty string.

koplas:
> Does not work for e.g. `rate` or `insecure`. Probably because the initial value is undefined or null but when you select and deselect a checkbox or enter a character and remove it from an input the new value is `false` or an empty string.

Does this still happen if you create a new source?

cintek:
> Does this still happen if you create a new source?

Yes.

Object before change:
```
{"id":3,"name":"Red Hat Product Security","url":"redhat.com","active":false,"rate":1,"slots":2,"strict_mode":true,"stats":{"downloading":0,"waiting":0},"ignore_patterns":[""]}
```

Object after I checked `insecure`:
```
{"id":3,"name":"Red Hat Product Security","url":"redhat.com","active":false,"rate":1,"slots":2,"strict_mode":true,"stats":{"downloading":0,"waiting":0},"ignore_patterns":[""],"insecure":true,"headers":[]}
```

Object after I unchecked `insecure`:
```
{"id":3,"name":"Red Hat Product Security","url":"redhat.com","active":false,"rate":1,"slots":2,"strict_mode":true,"stats":{"downloading":0,"waiting":0},"ignore_patterns":[""],"insecure":false,"headers":[]}
```

cintek (approved):
LGTM & works

PR #377, opened 2024-09-06
Add test csaf document

koplas:
This is needed to write useful integration tests.


bernhardreiter (changes requested):
The filename should adhere to https://docs.oasis-open.org/csaf/csaf/v2.0/os/csaf-v2.0-os.html#51-filename  otherwise put it in.

PR #386, opened 2024-09-10
Add initial statistics

koplas:
This pull request is only a mockup.


koplas:
The separate statistic menu is no longer required. This pull request is only a test of the chart library.

PR #389, opened 2024-09-11
Human usable age settings

luisa-beerboom:
And ensure that age can be emptied


client/src/lib/Sources/SourceCreator.svelte
```diff
@@ -93,6 +93,9 @@
 
   const saveAll = async () => {
     updateSourceForm();
+    if (source.age == "") {
```
koplas, line 96: nit: Use `===` for comparison, unless there is a specific reason for it.

client/src/lib/Sources/SourceForm.svelte
```diff
@@ -27,12 +29,80 @@
...
   let privateCert: FileList | undefined;
   let publicCert: FileList | undefined;
 
+  let ageUnit: AgeUnit;
+  let ageNumber: number | undefined;
+  let previousAgeNumber: number | undefined;
+
+  onMount(() => {
+    fillAgeDataFromSource(source);
+  });
```
koplas, line 85:
```Typescript
$: if (source) {
   // Do things that parse input from `source`
}
```
`onMount` will only be called only once. The source editor updates `source` after saving the source; refilling the age data may be required.

koplas (approved):
LGTM (and a couple of nits)

client/src/lib/Sources/SourceForm.svelte
luisa-beerboom: I haven't done this here because source.age would get overwritten every time the value in the inputs is changed.
Instead I made this dependant on the calling Component, from which `fillAgeDatafromSource` can be called, every time source was actually changed (see the changes in `SourceEditor.svelte`).

client/src/lib/Sources/SourceCreator.svelte
luisa-beerboom: Done

PR #392, opened 2024-09-11
Improve query designer

cintek:
Overall, should be working much better on smaller screens. Also, we use input fields for name and description of queries instead of a combination of input field *and* label. The combination lead to jumps of the elements when switching between display and edit mode and it makes code maintenance more complicated than necessary. The last improvement is that the error message when save failed is shown above the buttons so you don't have to scroll down to see it.


JanHoefelmeyer (approved):
LGTM & works

PR #395, opened 2024-09-12
Improve accessibility of buttons and checkboxes

cintek:
Targets/inputs must have a [minimum size](https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html) and there needs to be a [minimum contrast](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html) for colors. Gladly, Firefox has a tool to quickly be able to see if it's good enough.

Among others this solves the problem that if you e.g. don't hit the icon to delete a query you open the query designer accidentally which can be quite annoying.

In this PR I will only focus on these items:

- [ x ]  Size of checkboxes
- [ ] Size of icon buttons
- [ ] Contrast of icon buttons


JanHoefelmeyer (approved):
LGTM & works

PR #406, opened 2024-09-16
Validate roles with a type

s-l-teichmann:
When roles are processed from outside data they should be validated.
This PR does this.


koplas (approved):
LGTM

PR #407, opened 2024-09-16
feat: Restyle expanding cve column

### LINKED ISSUE #390: Search Page: Retain alternating color when opening CVEs
Expanded CVE always display a white background. In some cases this can cause white entries to follow up on each other, making it potentially difficult to distinguish different entries.

luisa-beerboom:
Instead of inserting a new row, add the additional data to the original row. Expand the clickable area to the entire text container to ensure ease of clicking.

Closes #390


JanHoefelmeyer (approved):
LGTM & works

PR #408, opened 2024-09-17
Small improvements for dashboard

cintek:
Several small improvements:

- Some additional attributes get a "special style" e.g. versions are shown as icon with a number.
- Use title for many parts of the tiles e.g. for title, tracking ID, SSVC etc.
- Improve wording of the message when there are no events
- Refactoring: Use a custom badge component to show the SSVC (used on the dashboard but also in the advisory single view)


JanHoefelmeyer (approved):
LGTM & works, still needs to be done:
https://github.com/ISDuBA/ISDuBA/issues/411

PR #410, opened 2024-09-17
Feat: Add github action for playwright tests, add script for playwright tests

JanHoefelmeyer:
Automates playwright tests via an action.

Also adds a script that will start keycloak, client and server and starts the playwright tests

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -1,0 +1,34 @@
+# This file is Free Software under the Apache-2.0 License
+# without warranty, see README.md and LICENSES/Apache-2.0.txt for details.
+#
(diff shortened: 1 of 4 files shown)


.github/workflows/playwright.yml
```diff
@@ -0,0 +1,34 @@
...
+        cd docs/scripts
+        ./setup.sh
+        ./playwright_tests.sh
+      shell: bash
+    - uses: actions/upload-artifact@v4
+      if: ${{ !cancelled() }}
+      with:
+        name: playwright-report
+        path: playwright-report/
+        retention-days: 30
```
koplas, line 34: This seems currently unused.

koplas (approved):
LGTM

PR #414, opened 2024-09-19
Dashboard (config)

### LINKED ISSUE #333: Activity layout problem (shortened, big double quotes, overlapping)
shows there layout problems:
 * the comment is unnecessarily shortened and does not give a clue that it continues like "Peter", so only "Pet" is seen.
 * the big double quotes are not closed and they are unnecessary anyway as  the placement and size of the contents of the comment already indicates what the text of the comment is
 * When resized to be small, the areas overlap, that should not happen, either the smaller onces get higher or they get completely folded below the "New Advisory" area. Anyway before overlapping the space to the left should be fully used.

seen with:
 ISDuBA: v0.2.1-alpha-274-g6469738

cintek:
- Fixes #333
- Makes it more clear which queries are the default ones in the query overview
- dashboard queries are only visible on the dashboard (not in the search)
- non-dashboard queries are only visible in the search
- `global` cannot be removed from default queries (will only be prevented by the backend but that's enough for now)

### DIFF
--- a/client/src/lib/Dashboard/AdvisoryQuery.svelte
+++ b/client/src/lib/Dashboard/AdvisoryQuery.svelte
@@ -92,18 +92,9 @@
             <Activity on:click={() => openDocument(doc)}>
               <div slot="top-left">
                 {#if doc.critical}
-                  <div>
-                    {#if doc.cvss_v3_score && doc.cvss_v3_score === doc.critical}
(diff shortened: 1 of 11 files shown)


JanHoefelmeyer (approved):
Contains improvements and works as intended.

PR #422, opened 2024-09-19
Add event view to table

### LINKED ISSUE #315: Search: Add events
For #313 we extend the queries with the type events. Since the `events` and the `documents` endpoint are built symmetrically it should be easily to extend the Search so users can search for events. What we could do is to add a button to switch to an event view (at 1. in the screenshot). Then we could show the appropriate queries (2. in the screenshot) so e.g. only event queries are visible when the user selected the event view.
search_header

Or we could show all queries at once and use some indicator to show the kind of query for each.

For this issue it's also important to test the full text search for events.

koplas:
Closes  #315

### DIFF
--- a/client/src/lib/Dashboard/EventQuery.svelte
+++ b/client/src/lib/Dashboard/EventQuery.svelte
@@ -16,7 +16,7 @@
   import { request } from "$lib/request";
   import { getErrorDetails, type ErrorDetails } from "$lib/Errors/error";
   import Activity from "./Activity.svelte";
-  import { Badge } from "flowbite-svelte";
+  import { Badge, Button } from "flowbite-svelte";
   import { push } from "svelte-spa-router";
   import { convertVectorToSSVCObject } from "$lib/Advisories/SSVC/SSVCCalculator";
   import { getRelativeTime } from "./activity";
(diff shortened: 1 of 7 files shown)


JanHoefelmeyer:
When using a non-event query while in the events tab of the search page, then the entire search table disappears.

We also really need to consider how to handle queries where apparently vital information is not wanted on the homepage.

Having time as NaN and Tracking ID as undefined is simply not nice.

Also the table for events is quite large and induces scrolling. Especially the timestamp should probably be abbreviated.

cintek:
> When using a non-event query while in the events tab of the search page, then the entire search table disappears.

I tried different combinations but I cannot reproduce this.

> We also really need to consider how to handle queries where apparently vital information is not wanted on the homepage.

This should be topic of a different issue/discussion.

> Having time as NaN and Tracking ID as undefined is simply not nice.

Another thing I cannot reproduce. Do you know for which advisory this happened or did it happen for all advisories?

> Also the table for events is quite large and induces scrolling. Especially the timestamp should probably be abbreviated.

The table filled with events doesn't seem larger than when filled with advisories so I think this is okay for now. Generally, there is potential for improvement when it comes to the layout of the table ;)

JanHoefelmeyer:
Seems I used a query that had no results? In that case, yeah, not a fault of this PR.

>> Having time as NaN and Tracking ID as undefined is simply not nice.
> Another thing I cannot reproduce. Do you know for which advisory this happened or did it happen for all advisories?
Simply create a query for the dashboard where you only want e.g. the actor. Then the card will show NaN days ago Tracking ID undefined

> The table filled with events doesn't seem larger than when filled with advisories so I think this is okay for now
It is but I agree that it's okay for now. But it's still an issue.

PR #428, opened 2024-09-23
fix: No longer give logging info that needs to be manually approved

JanHoefelmeyer:
Currently setup scripts won't always automatically run through but stop at the systemstatus of postgres. Unless there's a failure this is not  necessary, so removing it so it can run through smoothly.


cintek (approved):
LGTM

Issue #218, opened 2024-07-11
Diff: Box is getting resized when (de-)selecting documents


cintek:
To fix this we should limit the elements in the box in their width.

PR #437, opened 2024-09-23
Improve diff selection box

cintek:
- prevent changing width (Solves #218)
- work better on small screens
- remove error messages about failures regarding temporary documents when uploading a new document
- add tracking ID so it's easier for users to distinguish between the documents in the box

### DIFF
--- a/client/src/lib/Diff/DiffSelection.svelte
+++ b/client/src/lib/Diff/DiffSelection.svelte
@@ -238,7 +238,7 @@
 </script>
 
 <div
-  class="fixed bottom-0 left-1 right-1 flex flex-col items-end justify-center sm:left-auto md:right-20"
+  class="fixed bottom-0 left-1 right-1 flex flex-col items-end justify-center lg:left-auto lg:right-20"
 >
   <Button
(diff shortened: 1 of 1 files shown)


JanHoefelmeyer (changes requested):
Everything works as advertised, but the default now requires scrolling to select temporary documents for comparisons. 

Also the tooltip for the comparison button for temporay documents is broken declaring undefined as document, but this is consistent with main and not introduced with this PR

cintek:
> Everything works as advertised, but the default now requires scrolling to select temporary documents for comparisons. Maybe consider putting the + sign to the left, even if it is less consistent?

I moved the icons to the left. Maybe we should also do this in other areas like Search because there we also have this kind of problem when the table isn't fitting into the screen.

> Also the tooltip for the comparison button for temporay documents is broken declaring undefined as document, but this is consistent with main and not introduced with this PR

I fixed it anyway and made the tooltips of both icons consistent (before, both tried to show something different).

JanHoefelmeyer (approved):
LGTM & works

Issue #445, opened 2024-09-23
Query manager: Don't show edit form when user clones query


cintek:
## Situation now:
When you click on the clone button in the queries list

![queries1](https://github.com/user-attachments/assets/d2c57f4f-68e7-4ab1-9a1f-fa917aa49a6e)

you get to the query designer

![queries2](https://github.com/user-attachments/assets/4f9ef397-2863-477a-8b88-9b80c89ee984).

## Improvement:

When you click on the clone button in the queries list

![queries1](https://github.com/user-attachments/assets/d2c57f4f-68e7-4ab1-9a1f-fa917aa49a6e)

the query should be saved as a cloned variant and displayed in the list of personal queries.

![queries3](https://github.com/user-attachments/assets/291258f9-fccf-4275-a045-da98c240d016)

Then the user could open and edit it.

This makes #393 obsolete.

koplas:
Closed with #453.

PR #453, opened 2024-09-25
Improve clone behavior

cintek:
- Solves #445
- When an admin clones a global query the cloned query will still be global
- Cloned queries appear at the top of the list
- Scroll to the top of the table where the clone appears and animate the row of the clone so users see that the cloning process worked


JanHoefelmeyer (approved):
LGTM & works

PR #456, opened 2024-09-26
Allow to sort by multiple columns

### LINKED ISSUE #360: Search: Allow to sort by multiple columns
In the queries we allow to set multiple columns for the order of the results. Thus, the table in `Search` needs to support multiple sort factors.

koplas:
Closes #360


koplas:
The integration tests must be updated to be less fragile on table changes.

JanHoefelmeyer (approved):
Deselecting all order options seems to cause random orders, but that seems to be wanted then, since all orders were deselected. As such lgtm & works

PR #457, opened 2024-09-26
Preset age default in source create

luisa-beerboom:
Set MaxAge default for sources to 2 years and set ensured, the value would be pre-set in the SourceCreate


s-l-teichmann (changes requested):
`docs/isdubad-config.md` and `docs/example_isdubad.toml` also need adjustments.

cintek (approved):
LGTM & works

PR #465, opened 2024-09-30
Update flowbite

### LINKED ISSUE #438: Query-Manager: Grey out options that cannot be selected
Screenshot 2024-09-23 at 12-58-11 User defined queries

Here, I can't actually disable Test 1 being displayed on the dashboard since I'm only an auditor and not an admin. The coloring would suggest I can.

cintek:
Close #438


cintek:
Current problems:

- Checkbox doesn't have a `disabled` prop anymore (see https://github.com/themesberg/flowbite-svelte/issues/1454).
- FileUpload has a new required prop `value` which we didn't implement yet.

cintek:
I decided to first update only to 0.46.18. This is still fixing the issue but (hopefully) doesn't introduce new problems.

JanHoefelmeyer (approved):
LGTM & works

Issue #396, opened 2024-09-12
Queries: Add read-only view


cintek:
If you are no admin you have to clone a global query to see what is does. Max would like to see such things first before cloning the query so we should implement a read-only view of the `QueryDesigner`.

PR #468, opened 2024-09-30
Query designer: "read only" view

cintek:
Solves #396.

All users can now open any query. But if they are no admin they can only change the hide attribute of global queries.


koplas (approved):
LGTM & works

Issue #470, opened 2024-10-01
Update default query order


JanHoefelmeyer:
Current order is arbitrary, they should be listed in likelihood of what the user wants to see first.

JanHoefelmeyer:
Fixed via 471

PR #471, opened 2024-10-01
Fix: Improve default order of default stored queries

JanHoefelmeyer:
Since this is only a change in order of query-insertion I just changed the preexisting migration file. That so okay?

Background: If a user has multiple roles or can see multiple roles' queries (admin), then they still want to see the most relevant queries first. 

The general idea is the later in the workflow process the role is required the more important are the queries, which prevents bottlenecks at certain tasks.

Solves https://github.com/ISDuBA/ISDuBA/issues/470


s-l-teichmann:
This is not a valid migration. 020 is already in the database so you need a higher number. Furthermore you want to replace the old ones: DELETE the old ones first or use INSERT INTO ... ON CONFLICT.

s-l-teichmann (changes requested):
Please, make it a valid migration.

pkg/database/migrations/021-update_query_order.sql
```diff
@@ -0,0 +1,28 @@
...
+-- Software-Engineering: 2024 Intevation GmbH <https://intevation.de>
+
+
+DO $$
+DECLARE
+    default_definer constant varchar = 'system-default';
+    default_advisory_columns text array default Array['cvss_v3_score', 'cvss_v2_score', 'comments', 'critical', 'id', 'recent', 'versions', 'title', 'publisher', 'ssvc', 'state', 'tracking_id'];
+    default_event_columns text array default Array['cvss_v3_score', 'cvss_v2_score', 'comments', 'critical', 'id', 'title', 'publisher', 'ssvc', 'tracking_id', 'event', 'event_state', 'time', 'actor', 'comments_id'];
+BEGIN
+    DELETE FROM stored_queries WHERE definer = default_definer;
```
s-l-teichmann, line 16: You should really only delete what you want to replace. It could be possible that someone else has inserted data with the `system-default` user which would be destroyed by this, too.

JanHoefelmeyer:
Considering the hoops to jump through to make sure nothing ever breaks with this no matter what is configured this approach is not worth it.

s-l-teichmann:
Sorry, you only have to work more precisely. -> restored

pkg/database/migrations/021-update_query_order.sql
```diff
@@ -0,0 +1,278 @@
...
+-- Software-Engineering: 2024 Intevation GmbH <https://intevation.de>
+
+
+DO $$
+DECLARE
+    default_definer constant varchar = 'system-default';
+    default_advisory_columns character varying array default Array['cvss_v3_score', 'cvss_v2_score', 'comments', 'critical', 'id', 'recent', 'versions', 'title', 'publisher', 'ssvc', 'state', 'tracking_id'];
+    default_event_columns character varying array default Array['cvss_v3_score', 'cvss_v2_score', 'comments', 'critical', 'id', 'title', 'publisher', 'ssvc', 'tracking_id', 'event', 'event_state', 'time', 'actor', 'comments_id'];
+BEGIN
+    IF EXISTS (SELECT 1
```
s-l-teichmann, line 16: Looks like a hurtful looking variant of a simpler UPDATE statement.

s-l-teichmann (changes requested):

PR #472, opened 2024-10-01
Allow to delete and change workflow of multiple documents/advsiories

cintek:
As the title spoils this PR allows users to select multiple documents in the table and to delete them or change their workflow state. Later this multi-selection can be used to forward documents.

### DIFF
--- a/client/src/lib/Advisories/Version.svelte
+++ b/client/src/lib/Advisories/Version.svelte
@@ -34,7 +34,7 @@
   const navigateToVersion = (version: any) => {
     push(`/advisories/${publisherNamespace}/${trackingID}/documents/${version.id}`);
   };
-  const toggleDiffBoxActivated = () => {
+  const toggleToolboxActivated = () => {
     diffModeActivated = !diffModeActivated;
     if (diffModeActivated) {
       if (reversedAdvisoryVersions[0].version === selectedDocumentVersion) {
@@ -166,7 +166,7 @@
           {/each}
         {/if}
         {#if (appStore.isEditor() || appStore.isReviewer()) && advisoryVersions.length > 1}
-          <Button color="light" class="flex h-8 gap-x-2 px-3" on:click={toggleDiffBoxActivated}>
+          <Button color="light" class="flex h-8 gap-x-2 px-3" on:click={toggleToolboxActivated}>
             <i class="bx bx-transfer"></i>
             <span class="text-nowrap">{diffModeActivated ? "Hide" : "Show"} changes</span>
           </Button>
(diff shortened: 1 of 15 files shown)


cintek:
Currently, it looks like this:

multi-selection-marked

1. New checkboxes to (de-)select single or all documents. They should be on the left side of the table rows because on smaller screens they would not be visible if they would be on the right side. Furthermore, it is consistent because checkboxes are usually on the left side in other applications.
2. The "action buttons" to delete documents or (de-)select them for the diff are moved to the left. That is because the button for the diff-box also was moved to the left.
3. Next to the diff-button there are a buttons to delete or forward (just a disabled placeholder right now) all selected documents or change their state. These buttons should be near the checkboxes to it's easier for users to see what they can do with the selected documents. That is the reason why they are located on the left side and why the diff-button and the action buttons also should be on the left side.

JanHoefelmeyer (dismissed):
Source-Manager: "Upload Advisories" and the initial "Safe source" button now stretch to take up the entire side's width.

There seem to be 3 permanently disabled buttons next to the diff selection

cintek:
> Source-Manager: "Upload Advisories" and the initial "Safe source" button now stretch to take up the entire side's width.

I will look into that.
 
> There seem to be 3 permanently disabled buttons next to the diff selection

The first and the third one should be enabled as soon as you select at least one document. I will remove the second one for now (it was just like a preview how it would look like when there is a forwarding function).

JanHoefelmeyer (dismissed):
Binding the multi-operations to the diff-box seems unfortunate to me.

First, you'd expect these two to be related, it's not intuitive.

Second, you lose the multi-operations if you can't access diffs, e.g. as pure admin who can't utilize the multi-delete.

cintek:
> Second, you lose the multi-operations if you can't access diffs, e.g. as pure admin who can't utilize the multi-delete.

This was not intended and is a bug.

> First, you'd expect these two to be related, it's not intuitive.

I chose to implement some differences to make it easier for users to distinguish between them:

- The diff-button is aligned to the bottom of the screen, the multi-operation-buttons are not
- The diff-button contains a label, the multi-operation-buttons do not
- The multi-operation-buttons are smaller than the diff-button (even if you're not considering the label of the diff-button)

Furthermore, the place next to the diff-button seemed like an elegant solution to place the multi-operation-buttons so they don't look misplaced.

If this solution is not good enough a different idea is to place the multi-operation-buttons next to the paging controls:

multi-operation-buttons2

In this case I would prefer to put the select for the number of pages in front of the label for the select:

multi-operation-buttons

I didn't commit anything else yet since I first would like to be sure in which direction we are going.

JanHoefelmeyer (approved):
LGTM & works

Issue #480, opened 2024-10-08
Idling on Source-Manager tab will cause an error


JanHoefelmeyer:
Caused by GET http://localhost:5173/api/sources?stats=true

PR #481, opened 2024-10-08
Fix: Only pull source-stats when your role is allowed to fetch them (Editor, Source-Manager).

JanHoefelmeyer:
Solves https://github.com/ISDuBA/ISDuBA/issues/480


koplas (approved):
LGTM

Issue #466, opened 2024-09-30
Make Sources tab inaccessible for Auditor, Admin and Reviewer, Fix for Importer


JanHoefelmeyer:
Auditor, Admin and Reviewer can only see an error message in that tab because they shouldn't interact with it anyway. Makes no sense to let them try.

Importer will see an error message because they are not allowed to load the sources tab, but they need access due to the single-advisory-import. I'm partial to letting them see the sources similar to the Editor, but that might still warrant more thought.

PR #482, opened 2024-10-09
Fix: Hide add sources button from editor role (they are not allowed to add sources)

JanHoefelmeyer:
Remedies https://github.com/ISDuBA/ISDuBA/issues/466#issuecomment-2384900200


JanHoefelmeyer:
Also closes https://github.com/ISDuBA/ISDuBA/issues/466, since this is the last remaining issue from that issue

cintek (approved):
LGTM & works

PR #483, opened 2024-10-10
Add initial forwarder

koplas:
Adds an initial forwarder.

### DIFF
--- a/pkg/forwarder/forwarder.go
+++ b/pkg/forwarder/forwarder.go
@@ -87,7 +87,7 @@ func NewForwardManager(cfg *config.Forwarder, db *database.DB) *ForwardManager {
 			TLSClientConfig: &tlsConfig,
 		}
 
-		var headers http.Header
+		headers := http.Header{}
 		for _, header := range targetCfg.Header {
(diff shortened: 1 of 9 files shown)


s-l-teichmann (changes requested):
`docs/isdubad-config.md` needs a new `forwarder` section.

pkg/forwarder/forwarder.go
```diff
@@ -0,0 +1,389 @@
...
+			return err
+		}, 0,
+	); err != nil {
+		return documentIDs, err
+	}
+
+	return documentIDs, nil
+}
+
+func (fm *ForwardManager) fetchValidationStatus(ctx context.Context, documentID int64) validationStatus {
```
s-l-teichmann, line 303: Could be moved into the statement in loadDocument. To do so just create the SQL statement with the  `sb.WhereClause` from the SQLBuilder and not `.CreateQuery`

s-l-teichmann (changes requested):

JanHoefelmeyer (approved):
LGTM & works

s-l-teichmann (approved):

PR #484, opened 2024-10-10
Endpoints for import stats

s-l-teichmann:
This PR adds three new endpoints to the server:

1 - `GET /stats/imports/source/:id`
2 - `GET /stats/imports/feed/:id`
3 - `GET /stats/imports`

1 returns a list of accumulated imports for a source of given id. 2 does the same for a given feed. 3 for all sources.

The accumulation is parameterized with the query parameters `from`, `to` and `step`.
`from` and `to` are timestamps. Accepted formats  are `time.RFC3339`, `time.DateTime` and `time.DateOnly`
`step` is parsed as `time.Duration`. `from` defaults to now minus 3 days, `to` to now and `step` is the whole interval.

Special conditions are supported with the boolean query parameters `download_failed`, `filename_failed`, `schema_failed`, `remote_failed`, `checksum_failed` and `signature_failed`. By default all imports are returned.

Example usage against my system:
```
curl -s -D /dev/stderr --get \ 
--data-urlencode 'from=2024-01-02' \
--data-urlencode 'to=2024-10-11' \
--data-urlencode 'step=24h' \
--data-urlencode 'download_failed=false' \
http://localhost:8081/api/stats/imports/source/7 | jq .
```
results in:
```
[
  [
    "2024-09-19T00:00:00Z",
    501
  ],
  [
    "2024-09-23T00:00:00Z",
    4912
  ],
  [
    "2024-09-24T00:00:00Z",
    2300
  ],
  [
    "2024-10-10T00:00:00Z",
    6724
  ]
]
```

which is a list of two-element lists. The first items in the two-element lists are the start timestamp of a interval step. The second is the number of imports in this interval.

### DIFF
--- a/pkg/web/controller.go
+++ b/pkg/web/controller.go
@@ -171,5 +171,10 @@ func (c *Controller) Bind() http.Handler {
 	api.DELETE("/sources/feeds/:id", authSM, c.deleteFeed)
 	api.GET("/sources/feeds/:id/log", authSM, c.feedLog)
 
+	// Import stats
+	api.GET("/stats/imports/source/:id", authAll, c.importStatsSource)
+	api.GET("/stats/imports/feed/:id", authAll, c.importStatsFeed)
+	api.GET("/stats/imports", authAll, c.importStatsAllSources)
+
 	return r
 }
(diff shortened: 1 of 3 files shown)


koplas (approved):
LGTM

Issue #489, opened 2024-10-11
Source Manager: Scheduler unfair


s-l-teichmann:
Scenario: One Source with two Feeds.

1. Subscribe first feed
2. Downloader starts downloads from this queue till max downloads per source is reached.
3. Subscribe second feed.

Expected result: The downloader should download from both feeds fairly.
Observed result: The downloader only downloads from first feed till the feed queue is empty.

PR #490, opened 2024-10-11
Source Manager: Increase fairness in downloading from different feeds

s-l-teichmann:
This PR increases the fairness among the active feeds by shuffling them before picking advisories to download.

solve #489


koplas (approved):
LGTM

Issue #491, opened 2024-10-13
Source Manager: Downloader handles duplicates as failed


s-l-teichmann:
When a download succeeds but is detected as a duplicate while storing it to the database the download is treated as a failure which will result in downloading it again later.

Expected result: The downloaded document should be treated as **failed_duplicate** with
no reference to the copy which is already in the database. The `already` should add an entry
to the duplication URL. This will prevent the download loop.

PR #492, opened 2024-10-14
Source Manager: Store stats of duplicates and remember their download.

s-l-teichmann:
Solve #491

### DIFF
--- a/pkg/models/import.go
+++ b/pkg/models/import.go
@@ -336,16 +336,24 @@ func ImportDocumentData(
 	defer tx.Rollback(ctx)
 
 	const (
-		insertDoc     = `INSERT INTO documents (document, original) VALUES ($1, $2) RETURNING id`
-		insertLog     = `INSERT INTO events_log (event, state, actor, documents_id) VALUES ('import_document', 'new', $1, $2)`
-		queryText     = `SELECT id FROM unique_texts WHERE txt = $1`
-		insertText    = `INSERT INTO unique_texts (txt) VALUES ($1) RETURNING id`
-		insertDocText = `INSERT INTO documents_texts (documents_id, num, txt_id) VALUES ($1, $2, $3)`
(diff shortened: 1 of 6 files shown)


koplas (approved):
LGTM

Update:
I tested it with a source that already contains imported documents. I get the following error in the feed log:
`storing "https://security.access.redhat.com/data/csaf/v2/advisories/2024/rhsa-2024_4484.json" failed: already in database ERROR: current transaction is aborted, commands ignored until end of transaction block (SQLSTATE 25P02)`.

The downloads table does not contains a single entry where `duplicate_failed` equals true.

s-l-teichmann:
> LGTM
> 
> Update: I tested it with a source that already contains imported documents. I get the following error in the feed log: `storing "https://security.access.redhat.com/data/csaf/v2/advisories/2024/rhsa-2024_4484.json" failed: already in database ERROR: current transaction is aborted, commands ignored until end of transaction block (SQLSTATE 25P02)`.
> 
> The downloads table does not contains a single entry where `duplicate_failed` equals true.

Argh! Forgot to commit the transaction. Should be fixed now.

pkg/models/import.go
```diff
@@ -359,6 +359,9 @@ func ImportDocumentData(
 				if err2 := inTx(ctx, tx, 0, true); err2 != nil {
 					return 0, errors.Join(ErrAlreadyInDatabase, err2)
 				}
```
koplas, line 361: Does this statement need to be removed? I still get the same errors.

s-l-teichmann: No, the error canceled the whole transaction so it was not possible to write anything into the transaction afterwards. This is now circumvented by isolating the storing of the document with a save point.
This results in only rolling back to this and keep the transaction intact for further writes.

This should work now.

koplas:
It now works.

Issue #459, opened 2024-09-27
Search-Table CVE expanded status should be saved by CVE, not position


JanHoefelmeyer:
If two consecutive advisories on the search page have multiple cve's, then deleting the upper while its cves are expanded will cause the lower ones to expand their cve.

JanHoefelmeyer:
Here's an example: Between the upper and lower image, I only deleted the topmost advisory. The new topmost advisories' CVE is already expanded on the second image, even though that was not requested.
![Screenshot 2024-10-08 at 10-40-03 Search](https://github.com/user-attachments/assets/76521f54-a748-4989-98a1-ede91002d374)
![Screenshot 2024-10-08 at 10-40-14 Search](https://github.com/user-attachments/assets/abd38b5a-18ea-46c0-b83f-1bed135d7fea)

PR #505, opened 2024-10-16
Fix: reset expanded cves on reload

JanHoefelmeyer:
fixes https://github.com/ISDuBA/ISDuBA/issues/459


client/src/lib/Table/Table.svelte
```diff
@@ -309,6 +310,10 @@
     first();
   };
 
+  const setOpenRowNull = async () => {
+    openRow = null;
+  };
```
koplas, line 315: This function does not need to be async and can be inlined in `fetchData()`.

PR #506, opened 2024-10-16
Fix source header parsing

koplas:
The header field will currently reset in the SourceCreator if another field is edited. This pull request prevents this effect.


JanHoefelmeyer (approved):
LGTM & works

Issue #507, opened 2024-10-16
Dashboard/Queries: Switched order of queries when not admin


JanHoefelmeyer:
When not admin, then the global queries are displayed in the wrong order

JanHoefelmeyer:
Solved via #508 and #509

PR #508, opened 2024-10-16
fix: Reorder queries on original setup

JanHoefelmeyer:
part of https://github.com/ISDuBA/ISDuBA/issues/507


cintek (approved):
LGTM & works

PR #509, opened 2024-10-16
Fix: Clone default queries in opposite order since the first cloned will be the last in order after cloning

JanHoefelmeyer:
Second half of https://github.com/ISDuBA/ISDuBA/issues/507


cintek (changes requested):
Can you please squash the commits?

Issue #153, opened 2024-06-05
Home: Statistics


cintek:
To give the user the feeling that the application is working as expected and to give more information about the system we also want to display some statistics on the home page (only a portion of what will come with #48).

cintek:
Done by 49c13075158a45cd8ed2f615a1d4289f9865189f.

PR #511, opened 2024-10-16
feat: Add statistics to dashboard

cintek:
solves #153


cintek:
This is not perfect but we can refine the statistics in further PRs. For example the first label of the x axis leads to the gap on the left side of the chart.

Currently, the data is shown in 30 min steps. If we would take smaller steps this would mean that the bars would become really thin and almost not visible.

The chart itself is updated every 10 minutes.

dashboard-stats

I'd like to add that I started with the library [ApexCharts](https://apexcharts.com/) but in the end decided to switch to [Chart.js](https://www.chartjs.org/docs/latest/). Two reasons are:

- Labels in ApexCharts which are at the start or the end of the chart were just cut-off instead of adjusting the space (probably possible to fix but not a good default)
- If I was passing the same data to the chart (after a new fetch) it was changing the bars unexpectedly. Of course this is not helpful for a good usability.

JanHoefelmeyer (changes requested):
Timestamp is causing layout issues and is too precise for 2days, should be hours or similar.

If we want to show quickly whether the application is running via the statistics, then these statistics don't necessarily suffice since you may run into a situation where there are 0 advisories left to download, which is indistinguishable from having trouble downloading advisories.

Screenshot 2024-10-21 at 07-08-55 Dashboard

cintek:
Closing this in favor of #526.

PR #513, opened 2024-10-17
Change content of more-button on dashboard

cintek:
Since we are using the spreadsheet icon in the sidebar the user can easier see that the button will lead them to Search.


JanHoefelmeyer (approved):
LGTM & works

Issue #512, opened 2024-10-17
Search: Show comments in event table


cintek:
Currently, if you open the event table of Search you can see information about the related document and about the event but if the event is about a new comment the comment itself is not contained in the table. Since the comments are essential in this kind of events and we also show them on the dashboard for event queries we should show comments in the event table.

cintek:
Done by 12716d4deac4cc2cf6f1cdfe01ee89bc328faf60.

PR #522, opened 2024-10-18
feat: Show comments in event table

cintek:
Show comments in event table. Also, re-order columns because in an event table the information of the event itself are more important. Additionally, order events by time by default because users probably are more interested in what happend recently.

Solves #512.


JanHoefelmeyer (approved):
LGTM & works

PR #523, opened 2024-10-18
Keep track of manual imports

### LINKED ISSUE #518: Source Manager: Include manual imports and bulkimports into counting
Currently only imports from initiated from the source manager are going into counting stats.
The manual and bulk imports should be counted, too.

Idea: Create a source with too feeds (for manual and bulk) which is not modifiable via the
source manager. The id of the source should be 0. Every time a manual/bulk import is done the
respective feed could be used to store an entry in the downloads table.

koplas:
This adds a placeholder source containing all statistics for manual document imports.

TODO
- [x] Don't allow to edit the placeholder source
- [x] Handle the placeholder source in the frontend

Closes #518


cintek:
I would remove the values below "Domain/PMD", "Active" and "Loading/Queued" in the source overview. If you added some sources that would help to understand that the first source is something special and a string containing "invalid" sound like there is a problem that needs to be solved.
manual-imports-1

When it comes to the overview of the manual source I would remove "Domain/PMD" and the status at the top. Or can the status be something else than "OK"? In that case it's okay to keep the status. Again, I would remove the values below "URL" and "Loading/Queued".
manual-imports-2

JanHoefelmeyer (approved):
LGTM & works

Issue #48, opened 2024-03-13
Implement Statistics


JanHoefelmeyer:
Currently, the Statistics-Page is empty sans the word "Statistics". The functionality still needs to be implemented.

cintek:
Done

PR #526, opened 2024-10-18
Statistics everywhere

cintek:
Add statistics on the dashboard and in the tab `Statistics`. Solves #153 and #48.


cintek:
First progress: Reworked the component for the statistics so it can be used to show only imports (-> dashboard) or imports plus failures.

statistics-tab

cintek:
Now with controls to adjust the range:
statistics-tab-2

cintek:
Update: The same component is now able to display imports, import failures **and** CVEs.

cintek:
The labels on the x-axis of the chart are now depending on the whole range. For example if you select `Month` the chart is divided into calendar weeks. If the range is max. a day the labels contain only hours and minutes. This is addressing part of [requested changes](https://github.com/ISDuBA/ISDuBA/pull/511#pullrequestreview-2380934685) of another PR.

To prevent forgetting that issue mentioned by @JanHoefelmeyer I quote it here:
> If we want to show quickly whether the application is running via the statistics, then these statistics don't necessarily suffice since you may run into a situation where there are 0 advisories left to download, which is indistinguishable from having trouble downloading advisories.

Probably this won't be a problem when the application is used in a real scenario 'cause advisories will be published continuously and there will be more publishers and thus more sources. But we can discuss this further.

cintek:
An update of the statistics on the dashboard:
dashboard-stats

New: You can toggle to a table view of the component with the statistics (not enabled on the dashboard to keep it simple there)
stats-table

cintek:
@JanHoefelmeyer can you please test again? I improved the colors used in the statistics a little bit but there are limits. On a small screen and a high "resolution" (small steps in a long time range) the bars will always be very thin if many datasets are shown. But of course the user can change the time range and hide specific datasets so it will be easier to distinguish the visible ones.

JanHoefelmeyer (approved):
LGTM & works

PR #527, opened 2024-10-19
Stats: count documents and advisories

s-l-teichmann:
This PR adds an endpoint

- `GET` `/api/stats/totals`

to count documents and advisories.

with the optional query parameters `time` and `imports`. `time` defaults to now and may be
of format `time.RFC3339`, `time.DateTime` or `time.DateOnly`. `imports` defaults to false.

`time` determines the point in time when the counters should be probed.
`imports` determines if the documents should be counted by their import time stamps or
their current release dates. Documents with a current release date in the future are assumed to be imported by now.

Example usage from my system:

```
curl -s -D /dev/stderr --get \                                                             stats_total?
--data-urlencode 'time=2024-10-18' \
--data-urlencode 'imports=false' \
http://localhost:8081/api/stats/totals | jq .
```

results in
```
{
  "advisories": 24286,
  "documents": 24570
}
```

### DIFF
--- a/pkg/web/controller.go
+++ b/pkg/web/controller.go
@@ -178,6 +178,7 @@ func (c *Controller) Bind() http.Handler {
 	api.GET("/stats/cve/source/:id", authAll, c.cveStatsSource)
 	api.GET("/stats/cve/feed/:id", authAll, c.cveStatsFeed)
 	api.GET("/stats/cve", authAll, c.cveStatsAllSources)
+	api.GET("/stats/totals", authAll, c.statsTotal)
 
 	return r
 }
--- a/pkg/web/totalstats.go
+++ b/pkg/web/totalstats.go
@@ -10,12 +10,69 @@
 package web
 
 import (
+	"context"
+	"errors"
+	"log/slog"
 	"net/http"
+	"strconv"
+	"time"
 
 	"github.com/gin-gonic/gin"
+	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgxpool"
 )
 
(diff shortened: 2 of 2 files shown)


s-l-teichmann:
> Calling it while downloading advisories may cause problems:
> 
> curl -s -D /dev/stderr --get -H "Authorization: Bearer $TOKEN" --data-urlencode 'time=2024-10-18' --data-urlencode 'imports=false' http://localhost:5173/api/stats/totals | jq . HTTP/1.1 200 OK Access-Control-Allow-Origin: * content-type: application/json; charset=utf-8 x-request-id: 8acd912b-ce12-47dd-97c4-2bee1a6e96d4 date: Mon, 21 Oct 2024 05:38:08 GMT content-length: 34 connection: close
> 
> { "advisories": 897, "documents": 896 }
> 
> This is on a source that offers exactly 1 advisory per document as far as I can tell.
> 
> Subsequent calls do not have this anomaly:
> 
> HTTP/1.1 200 OK Access-Control-Allow-Origin: * content-type: application/json; charset=utf-8 x-request-id: 6fe2c702-7d6e-4b69-96fb-dfb859c23f53 date: Mon, 21 Oct 2024 05:40:33 GMT content-length: 36 connection: close
> 
> { "advisories": 1909, "documents": 1909 }
> 
> The timing seems to have to be quite precise since I couldn't reproduce it, but it's worth mentioning. Other than that, LGTM & works.

Mhmm ... both the import and the counting run in transactions. The import transaction includes the updating of the advisories. So this is really suspicious. Maybe its your test or the transaction are not isolated from each other. I would have a look at this when we are going replace the (publisher, tracking_id) document/advisory joins.

Issue #524, opened 2024-10-18
Search: Handle sorting in case sort value does not exist


JanHoefelmeyer:
E.g. sort events by time and then switch to documents to get a 400 due to time not existing as field for documents.

PR #533, opened 2024-10-21
Fix: Filter out invalid order states when switching query types

JanHoefelmeyer:
closes https://github.com/ISDuBA/ISDuBA/issues/524


koplas (changes requested):

koplas (approved):
LGTM

PR #535, opened 2024-10-21
Fix header handling

### LINKED ISSUE #521: Source-Manager: Not possible to delete HTTP headers
- Create source and enter some HTTP headers
- Open the source delete the headers and click on "Save source"

Expected: The headers should be removed permanently.

Result: The removed headers appear again.

When you navigate to another page and come back you can verify that the headers were *not* removed.

koplas:
Closes #521


JanHoefelmeyer (approved):
LGTM & works

PR #538, opened 2024-10-24
Statistics: Probe imports in intervals

s-l-teichmann:
This PR modifies the endpoint

- `GET` `/api/stats/totals`

in a way to allow probing in intervals. They query parameter `time` is not supported any longer.
Instead the parameters `from`, `to` and `step` are introduced. `from` defaults to now.
if `to` is not given it is assumed to be the same value as `from`. If `step` is not given it will
be set to the length of the the interval from `from` to `to`.

Example from my system:
```
curl -s -D /dev/stderr --get \ 
--data-urlencode 'from=2024-10-22' \
--data-urlencode 'to=2024-10-24' \
--data-urlencode 'step=24h' \     
--data-urlencode 'imports=false' \           
http://localhost:8081/api/stats/totals | jq .
```

results in
```
[
  [
    "2024-10-22T00:00:00Z",
    25525,
    25194
  ],
  [
    "2024-10-23T00:00:00Z",
    25570,
    25211
  ],
  [
    "2024-10-24T00:00:00Z",
    25588,
    25227
  ]
]
```

which is a list of 3-element lists. The first element in a 3-element list is the probe time, second and third are document ans advisory counts.

BTW: There is a limitation to the length of the list to ~32.000 entries. This due to SQL request strategy I have  chosen (batch queries). This could be done differently resulting in potential weaker performance.

### DIFF
--- a/pkg/web/importstats.go
+++ b/pkg/web/importstats.go
@@ -91,6 +91,8 @@ func (c *Controller) criticalStatsFeed(ctx *gin.Context) {
 	c.importStatsFeedTmpl(ctx, selectCriticalSQL, collectCritcalBuckets)
 }
 
+const importStatsDefaultInterval = 3 * 24 * time.Hour
+
 func (c *Controller) importStatsSourceTmpl(
 	ctx *gin.Context,
 	sqlTmpl string,
@@ -100,7 +102,7 @@ func (c *Controller) importStatsSourceTmpl(
 	if !ok {
 		return
 	}
-	from, to, step, ok := importStatsInterval(ctx)
+	from, to, step, ok := importStatsInterval(ctx, importStatsDefaultInterval)
 	if !ok {
 		return
 	}
@@ -122,7 +124,7 @@ func (c *Controller) importStatsAllSourcesTmpl(
 	sqlTmpl string,
 	collectBuckets func(rows pgx.Rows) ([][]any, error),
 ) {
-	from, to, step, ok := importStatsInterval(ctx)
+	from, to, step, ok := importStatsInterval(ctx, importStatsDefaultInterval)
 	if !ok {
 		return
 	}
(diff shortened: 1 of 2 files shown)


JanHoefelmeyer:
Clarification regarding the assumed values, for whom testing showed me the following behaviour:

- If *from* or *to* are not given, their value is set to 'now'. 
- If *to* is set to an earlier point in time than *from*, their values will switch, meaning *from* is treated as *to* and vice versa.
- setting *step* to any valid value has no effect if *from* and *to* are equal. 
- If *step* is greater than the difference between *from* and *to*, then only the value from *from* is returned. 
- The max value for *step* is '2562047h' or slightly over 292 years
- If *step* is not set, it will be set to the minimum of the difference between *from* and *to* and it's max value.
- Any invalid value in *from*, *to* and *step* will cause a syntax error
- Too many *steps* will cause an error "too many steps"

If that is working as intended, then it works

s-l-teichmann:
> Clarification regarding the assumed values, for whom testing showed me the following behaviour:
> 
>     * If _from_ or _to_ are not given, their value is set to 'now'.

yes

>     * If _to_ is set to an earlier point in time than _from_, their values will switch, meaning _from_ is treated as _to_ and vice versa.

yes
 
>     * setting _step_ to any valid value has no effect if _from_ and _to_ are equal.

yes

>     * If _step_ is greater than the difference between _from_ and _to_, then only the value from _from_ is returned.

yes

>     * The max value for _step_ is '2562047h' or slightly over 292 years

maybe (haven't calculated the actual value). If this will cause an issue, we all will be dead then.

>     * If _step_ is not set, it will be set to the minimum of the difference between _from_ and _to_ and it's max value.

yes
>     * Any invalid value in _from_, _to_ and _step_ will cause a syntax error

yes

>     * Too many _steps_ will cause an error "too many steps"

yes
 
> If that is working as intended, then it works

JanHoefelmeyer (approved):
works & LGTM

PR #539, opened 2024-10-24
Fix: Don't allow to forward unless target is specified

JanHoefelmeyer:
Currently you can try to forward without selecting a forwarding target. This PR hides the button until a target is selected


koplas (approved):
LGTM

PR #549, opened 2024-10-28
Change max_age to default_age

koplas:


JanHoefelmeyer (approved):
LGTM & works. Please look over my changes before merging

client/src/lib/Sources/SourceForm.svelte
```diff
@@ -302,7 +302,7 @@
               class="rounded-none rounded-l-lg"
               type="number"
               min="0"
-              placeholder={agePlaceholder.toString()}
+              placeholder=0
```
koplas: This seems to be wrong. Is there a specific reason for this change?

JanHoefelmeyer: The placeholder value appears exactly when the internal value is "" or empty, achieved by actively using 0. The old code would then show the default maximum age as value.

For example, assume the default maximum age is 1 week, but you manually set the age to unlimited, then the display would default to 1 week, even though its using the 0 value internally.

While the 1 week default value is distinguishable from an actual 1 week value because the number is greyed out, it's very likely to lead to some confusion. Since the value is only ever seen when choosing 0, there should be no problem when setting it to 0.

Note that this does not affect the preset value when first creating a source, which is still taken from the server and is not greyed out.

PR #550, opened 2024-10-28
Aggregator support

s-l-teichmann:
This PR add support for aggregators.


pkg/web/aggregators.go
```diff
@@ -0,0 +1,209 @@
...
+	if !ok {
+		return
+	}
+	const sql = `DELETE FROM aggregators WHERE id = $1`
+	var deleted bool
+	if err := c.db.Run(
+		ctx.Request.Context(),
+		func(rctx context.Context, conn *pgxpool.Conn) error {
+			tag, err := conn.Exec(rctx, sql, id)
+			deleted = tag.RowsAffected() > 1
```
koplas, line 196: Shouldn't the check be >= 1?

JanHoefelmeyer (changes requested):
Manual imports has a permanent "Attention here" exclamation sign that cannot be resolved. Might be related to https://github.com/ISDuBA/ISDuBA/issues/574 ?

There's no reason to be able to mark a source as done while creating it when creating it will automatically mark it as done.

cintek:
> Manual imports has a permanent "Attention here" exclamation sign that cannot be resolved. Might be related to #574 ?
> 
> There's no reason to be able to mark a source as done while creating it when creating it will automatically mark it as done.

All new sources have `attention` set to `true`. That is the reason for both of your observed cases. I think that this should be different, too.

PR #551, opened 2024-10-28
feat: Move the forward button to the bottom-right

### LINKED ISSUE #544: Sensible location for forward button

luisa-beerboom:
Closes #544


JanHoefelmeyer (changes requested):
I think there needs to be a better solution regarding having no rights to access the comment section etc.

In that case the forward button will simply be located at the bottom right of the page, which may require quite some scrolling in case of very long branches tabs or similar.
Screenshot 2024-10-28 at 19-18-41 RHSA-2023 0632

PR #552, opened 2024-10-28
feat: improve search-checkbox usability

### LINKED ISSUE #536: Search: Checkbox and diff-selection columns should not link to advisory detail
Using the checkboxes (or the diff selection column) in the search list can be quite frustrating, as one misclick (something like "accidentally clicking 2px to the right of the checkbox") will be read as wanting to see the advisory detail page of that row, leading to navigation.
This is especially bad if multiple rows are to be selected, as one can lose one's entire selection just by missing the checkbox on the last click. As a result, users have to pay an undue amount of attention or risk some annoyance if they try to use this feature.

We should change things so that clicking into the back of these columns does either one of:
- activating the checkbox as if it had been clicked, or
- nothing, forcing the user to try again

luisa-beerboom:
Closes #536 

The table on the search page now only links to the advisories on the fields without clickable features. This means:
- the entire checkbox column
- the entire diff/delete selection column
- the cve field for any advisory which has multiple

Additionally:
Removed code-based routing logic, by using <a> tags instead.

### DIFF
--- a/client/src/lib/Table/Table.svelte
+++ b/client/src/lib/Table/Table.svelte
@@ -519,7 +519,7 @@
         </TableHead>
         <TableBody>
           {#each documents as item, i}
-            <tr>
+            <tr class={i % 2 == 1 ? "bg-white hover:bg-gray-200" : "bg-gray-100 hover:bg-gray-200"}>
               {#if isMultiSelectionAllowed}
                 <TableBodyCell tdClass="px-0">
                   <CCheckbox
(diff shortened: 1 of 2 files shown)


JanHoefelmeyer (changes requested):
Screenshot 2024-10-30 at 10-14-24 Search

I think reimplementing the coloring of rows would be beneficial, see:
Screenshot 2024-10-30 at 10-14-04 Search

JanHoefelmeyer (approved):
LGTM & works

PR #556, opened 2024-10-29
Upgrade to Keycloak 26

### LINKED ISSUE #485: consider moving to keycloak 26
https://www.keycloak.org/2024/10/keycloak-2600-released

koplas:
Closes #485


bernhardreiter:
Hmm looking at the tests, did the unittests run? (And why do we have to runs for playwright and REUSE, can be maybe prevent one of them?)

koplas:
REUSE is always required to run. I don't want to disable the playwright run, because the playwright run is useful to test Keycloak and playwright makes usage of the setup scripts. Unittests only run if something of the client has changed. I consider the current behavior as correct and don't want to change it.

JanHoefelmeyer (approved):
LGTM & works

PR #562, opened 2024-10-30
Diff - Show loading indicators

### LINKED ISSUE #217: Diff: Show some indicator when user selected a document
This is important when users have a low internet speed. When you select a document from the table the diff icon is gray but you see no changes in the diff box:
diff_selected_doc

Background: The data in the table depends on the query that is used. It can be that we don't have every information we want to show in the diff box. That's why we only store the ID of the selected document in the Svelte store. Then we load the document separately in the diff box. Thus, there is some delay until the information is shown in the box.

Idea: When a document is selected for the diff show some elements right away in the box: The `x`-button which is used to remove a document from the selection (that is now already visible *after* loading the document) and maybe the ID of the document. We also need a loading indicator.
When we show the `x`-button right away we can also allow the users to abort loading of the document.

cintek:
- when a user selects a document a loading indicator is displayed in the diff box
- when the user then clicks on "Compare" they get to the diff page where another loading indicator is displayed until the diff is ready
- remove redundant requests
- some refactoring

Fixes #217.

### DIFF
--- a/client/src/lib/Diff/Diff.svelte
+++ b/client/src/lib/Diff/Diff.svelte
@@ -9,7 +9,7 @@
 -->
 
 <script lang="ts">
-  import { Accordion, AccordionItem, Button, ButtonGroup, Label } from "flowbite-svelte";
+  import { Accordion, AccordionItem, Button, ButtonGroup, Label, Spinner } from "flowbite-svelte";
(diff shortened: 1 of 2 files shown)


JanHoefelmeyer (approved):
LGTM & works

PR #566, opened 2024-10-30
feat: Ensure the general area has word breaking

### LINKED ISSUE #215: Document view: Text can overlap comment box
If e.g. "Publisher issuing authority" and "Aggregate severity namespace", which are now next to each other, are too long, then it can take up nearly the entire site's width. In this case, the comment box will simply cut it off.

luisa-beerboom:
Closes #215 

Now using grid instead of table in that place since tables have a tendency to expand beyond the parent div's area.


JanHoefelmeyer (changes requested):
image

Could this be adjusted for the other problematic section too?

JanHoefelmeyer (approved):
LGTM & works

PR #579, opened 2024-11-06
feat: Remove multiple tooltips that didn't add new information

### LINKED ISSUE #340: Document View: improve tooltips
The tools tips for the view of a single document should all be checked for usefulness.

To 
  * remove tooltips which say the same as a string that is already fully shown.
  * remove tooltips that make no sense, like showing the category when hovering over a detail.

Less toolstips are helpful as they do not block the view on the data below.

luisa-beerboom:
Closes #340 

Places:
- Collapsible: Tooltips can now be set via the 'title' attribute, which is currently unused
- Overview doesn't show the product title tooltips any more
- SSVC Calculator had multiple buttons where the tooltip had the exact same content as the button, removed those tooltips

### DIFF
--- a/client/src/lib/Advisories/CSAFWebview/Collapsible.svelte
+++ b/client/src/lib/Advisories/CSAFWebview/Collapsible.svelte
@@ -10,6 +10,7 @@
 
 <script lang="ts">
   export let header: string;
+  export let title: string | undefined = undefined;
   export let open = false;
   export let level = 2;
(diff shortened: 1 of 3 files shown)


cintek (changes requested):
I suggest to also remove the tooltips of the buttons at the bottom right when you started the evaluation as they don't contain much more information than the labels of the buttons:
image

PR #581, opened 2024-11-06
Use custom file inputs

### LINKED ISSUE #575: Issues with Fileinput in `Sources`
There are at least two issues with the `Fileinput` in `Sources`:

- If you selected a file and click on the button with the "x" the filename (see 1. in the screenshot) won't be removed. For the user it seems like the button isn't working.
- When no file was selected the button to remove the selection is active (see 2. in the screenshot) but it shouldn't.

certs

cintek:
Closes #575 and finally the button has the primary color so it is more consistent.

I close #580 because it was more a quick fix like @koplas told me.

### DIFF
--- a/client/src/lib/Components/CFileinput.svelte
+++ b/client/src/lib/Components/CFileinput.svelte
@@ -18,6 +18,7 @@
   export let files: FileList | undefined = undefined;
   export let id: string;
   export let multiple = true;
+  export let oldFile: string | null | undefined = undefined;
   export let titleClearButton = "";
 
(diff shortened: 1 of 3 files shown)


koplas (changes requested):
The reset button does not work. To reproduce:
- Select file
- Click on X button next to the file picker
- Click `Save source`

The result is that the file is still uploaded.

If we implement a custom file input, we should display if a cert file has already been uploaded and allow the removal of the cert file by clicking on the X button.

koplas (approved):
LGTM

PR #583, opened 2024-11-06
fix: return valid boundaries if slot value is out of range

### LINKED ISSUE #520: Improve error message if source-manager manual values are out of range
The error message for e.g. slots will state "The request sent could not be understood. updates failed: slot value ot ouf range". It should be clearer that this is configured and intended and not a parsing problem or technical limitation.

s-l-teichmann:
resolves #520

```
curl -s -D /dev/stderr -X PUT \                                   slot_range?
--data 'slots=1000' \
http://localhost:8081/api/sources/3 | jq .
HTTP/1.1 400 Bad Request
Content-Type: application/json; charset=utf-8
X-Request-Id: 58b1ee4a-5fb4-4198-a754-2f01e0330840
Date: Wed, 06 Nov 2024 19:35:28 GMT
Content-Length: 77

{
  "error": "updates failed: slot value out of range range: 1000 not in [1, 2]"
}
```

### DIFF
--- a/pkg/sources/manager.go
+++ b/pkg/sources/manager.go
@@ -1154,9 +1154,10 @@ func (su *SourceUpdater) UpdateSlots(slots *int) error {
 	if slots != nil && su.updatable.slots != nil && *slots == *su.updatable.slots {
 		return nil
 	}
-	if slots != nil && (*slots < 1 ||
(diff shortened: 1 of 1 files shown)


koplas (approved):
LGTM & works

PR #587, opened 2024-11-07
Simple docs

JanHoefelmeyer:


bernhardreiter:
The texts in general have too many redundancies and fillers that state the obvious.

PR #588, opened 2024-11-07
Invalidate ETags/last_modified headers if we need a feed index re-read.

### LINKED ISSUE #519: Source-Manager: Increasing age-range should trigger reevaluation of feeds
For example, trying to set age range to 2 years may yield ~ 8k documents to download. Setting it down to 2 months then yields 2k. Setting it back up  to 2 years I'd expect it to go back to ~ 8k, but it will simply be still 2k. This is especially noticable when setting the time to 0, which will stop any download even if setting it back up again.

The download wouldn't resume until 15 minutes later after another feed refresh.

s-l-teichmann:
resolves #519

This PR invalidates the ETags/last_modifled headers to re-read the feed index
if we need it (e.g. set max release date filter).

### DIFF
--- a/pkg/sources/manager.go
+++ b/pkg/sources/manager.go
@@ -1116,6 +1116,16 @@ func placeholders(n int) string {
 type SourceUpdater struct {
 	updater[*source]
 	clientCertUpdated bool
+	doBackgroundPing  bool
+}
+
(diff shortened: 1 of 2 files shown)


koplas (approved):
LGTM & works

PR #589, opened 2024-11-07
feat: Make selected Advisory Product/CVE blink after scrolling

### LINKED ISSUE #572: Vulnerabilities Overview: Make more obvious why and where you jumped to
Clicking on a vulnerability will jump you to it, but there is no indicator to what you jumped to so you have to keep searching. 
E.g. after clicking on a vulnerability I jump to this and have no idea what I jumped to:
image

luisa-beerboom:
Closes #572 

Added a new animation for this: Area will be grey and fade back to background colour within 4s

### DIFF
--- a/client/src/app.pcss
+++ b/client/src/app.pcss
@@ -30,3 +30,22 @@
     opacity: 1;
   }
 }
+
+/*
+Webview styles
+*/
+
+@keyframes blink {
+  from {
+    background-color: gray;
+  }
+  to {
+    background-color: inherit;
+  }
+}
+
+.blink {
+  animation-name: blink;
+  animation-duration: 4s;
+  border-radius: 0.5rem;
+}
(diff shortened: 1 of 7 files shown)


client/src/app.pcss
```diff
@@ -30,3 +30,22 @@
...
+  to {
+    background-color: inherit;
+  }
+}
+
+.blink {
+  animation-name: blink;
+  animation-duration: 4s;
+  border-radius: 0.5rem;
+}
```
koplas, line 51: We also have a blink animation in `SourceForm.svelte:412-429`. If you like you can replace the animation in SourceForm with this animation. I prefer the animation you defined more and this would avoid duplication.

koplas (approved):
LGTM & works

client/src/app.pcss
luisa-beerboom: Done

PR #590, opened 2024-11-07
Diff layout improvements

### LINKED ISSUE #101: JSON Diff: Use table for changes in the same path
Sometimes we have multiple changes in a path like `/document/references/`. To prevent duplicate keys (and thus lower mental load) we could organize the information in a different way and use tables.

cintek:
TODO:

- [ ]  summarize changes that affect the same part of the document (e.g. all additions in `remediations`)
- [ ] make formatting of lists more compact
- [ ] use anchor element for URLs

Closes #101


cintek:
With 0bad412260d22eb1ab7c454e464d62e51cc057f5 the application also summarizes `remove` changes. Problem is that it first takes a look at the path to see if there is something that has the potential to be summarized. If that is the case it looks into the content of a change to be sure if a summary is possible. But to do this it has to load many changes which may take some time.

So, there are two possibilities:

1. Only summarize `add` changes -> this decreases consistency between `add` and `remove` changes further
2. Summarize both -> long loading time for `remove` changes so it takes some time until the user can see some content in the accordion for `remove` changes

PR #591, opened 2024-11-07
Extend aggregator viewer

koplas:
- [x] Display what source or feeds are already subscribed
- [x] Allow to jump to already configured sources


cintek (approved):
LGTM & works

PR #599, opened 2024-11-10
Aggregator: Join aggregator source and subscripted sources by canonical URLs

s-l-teichmann:
This PR joins the provider URLs from the aggregator with the subscripted ones by their canonical representation take from the PMDs.


koplas (approved):
LGTM & works

PR #615, opened 2024-11-12
Allow to delete source client certificates

### LINKED ISSUE #609: Can't remove uploaded client certs in source manager
Removing the client certificates after they have been uploaded is currently impossible. Clicking on the `X` button and saving the source should reset these fields.

koplas:
Closes #609


cintek (approved):
LGTM & works

PR #617, opened 2024-11-12
fix: Make statistics text colors coherent with dark mode activation

luisa-beerboom:
…after reload

Moved dark mode recognition code from the button to `+page`.


koplas (approved):
LGTM & works

Issue #585, opened 2024-11-06
Improve reporting on import errors


tschmidtb51:
Currently, I see some import errors in the statistic page, but it is hard to figure out which source is the reason (especially, if you have multiple sources).
![grafik](https://github.com/user-attachments/assets/9689db82-6e95-4e8b-9ab5-3c8056132649)

Moreover, on a single source, it is hard to figure out which feed is the reason if you have multiple feeds.

Analyzing the root cause might also be a problem as the pagination of the logs covers only 100 entries so that `Ctrl+F` might not help in the browser - especially, if the problem occured two or three days ago (weekend...)

bernhardreiter:
@luisa-beerboom can you add 1000 here as posssiblity and a download button?

tschmidtb51:
@bernhardreiter I like the download button. Maybe an "all" option would also be helpful? Also sorting the columns might be helpful.

luisa-beerboom:
> @luisa-beerboom can you add 1000 here as posssiblity and a download button?

What should the download button download?

tschmidtb51:
> > @luisa-beerboom can you add 1000 here as posssiblity and a download button?
> 
> What should the download button download?

The entire log file.

bernhardreiter:
> The entire log file.

(Which could be too much, so it would not be perfect)

bernhardreiter:
There is another improvement to the underlying use case in #621

bernhardreiter:
Solved good enough with #694 #708

PR #618, opened 2024-11-12
feat: Improve feed log information access

luisa-beerboom:
See #585 

Increase upper limit on page pagination to 10000
Downloadable log file per feed (json file with feed name title for now -> Is there any preferred format?)


client/src/lib/Sources/FeedLogViewer.svelte
```diff
@@ -96,6 +97,32 @@
     }
   };
 
+  const downloadFeedLogs = async () => {
+    if (!feed) {
```
cintek, line 101: Maybe nit-picking but you could summarize the first two if-clauses to make the code a little more compact.

cintek (approved):
LGTM & works

client/src/lib/Sources/FeedLogViewer.svelte
cintek: Thanks :)

PR #619, opened 2024-11-12
Force aggregator schema validation

koplas:
After upgrading the gocsaf/csaf dependency, it is possible to validate the aggregator schema correctly. This pull request removes the `validate` parameter from the aggregator proxy endpoint.


s-l-teichmann (approved):
LGTM & works

PR #625, opened 2024-11-13
Refine dark mode

### LINKED ISSUE #620: Dark mode: Refinement
This is an issue to collect issues with the dark mode where it needs some refinement.

What I found so far:

When a dashboard query gives no results the text has poor contrast (notice that this is also the case for event queries which are not shown in this screenshot):
dashboard_queries

If there are tiles on the dashboard the text/icons at the bottom also have too poor contrast:
tile

If you hover over the button to show the diagram of the statistics component the icon is not visible:
statistics_hover

When you open the diff box at the bottom of `Search` and selected two documents the text at the bottom has poor contrast:
diff_selection

When you click on compare in the diff box and get to the diff the path of the changes of "Removed" have poor contrast:
diff_removed

In the diff component when you click on "Inline" the colors are not adjusted yet:
diff_inline

BTW: When I'm saying poor contrast that means that I tested it with the accessibility dev tool of Firefox which tells if a contrast is sufficient or not.

luisa-beerboom:
Closes #620

### DIFF
--- a/client/src/lib/Advisories/History.svelte/History.svelte
+++ b/client/src/lib/Advisories/History.svelte/History.svelte
@@ -86,7 +86,7 @@
     </TableBody>
   </Table>
   {#if historyEntries.length === 0}
-    <span class="ml-auto mr-auto text-gray-400 dark:text-gray-600"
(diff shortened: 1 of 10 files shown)


cintek (approved):
LGTM & works

PR #627, opened 2024-11-13
feat: More specific names for HTTP headers in source manager

### LINKED ISSUE #626: Improve specifying request headers by using more specific name
Feedback from one user was, that it was unclear if additional header fields in the sources would construct the header or need a `:` to work.

My suggestion is to rename the label to be more HTTP specific.

https://datatracker.ietf.org/doc/html/rfc9110#name-fields

> HTTP uses "fields" to provide data

> Field sections are composed of any number of "field lines", each with a "field name"
 [..] "field value"

So I suggest to rename `key` to `field name` and `value` to `field value` in the interface. 

By getting closer to the RFC terminology and being more specific, we lower the chance of missunderstanding.

luisa-beerboom:
Closes #626


koplas:
I don't know if this is more specific than the current state. @bernhardreiter

koplas (approved):
LGTM

PR #628, opened 2024-11-13
Ignore empty patterns in regexes in download filtering

s-l-teichmann:
Do not try to use empty strings for filtering.


koplas (approved):
LGTM & works

PR #630, opened 2024-11-13
Improve layout of aggregators

cintek:
Among others makes it easier to distinguish between aggregators and providers/feeds.


koplas (approved):
LGTM & works

PR #633, opened 2024-11-13
Dashboard improvements

cintek:
This PR contains small improvements for the dashboard:

1. Increase space between queries and statistics -> consistent to space between queries
2. Increase contrast of the text "No source changes found." in dark mode
3. Ensure button to show more is always at bottom (see screenshot how it was before the change if there was only one result) -> More consistency 'cause this button is already at the bottom when there are more results
query_more


koplas (approved):
LGTM & works

PR #635, opened 2024-11-13
More query parameters to filter feed logs.

### LINKED ISSUE #621: Enable deeper log analysis of source fetching failures via API
A source admin (Peter) wants to find out where the download failures in the diagrams and counts originate.
He wants to have the "full" log.

To help with this, an endpoint that allows to get the log entries for all feeds is useful, so Peter can use script.

In addition we document and example how Peter can get an access token from the keycloak to use the script.

s-l-teichmann:
resolves #621 

This PR adds a new endpoint to access the logs to all feeds:

* GET `/api/sources/feeds/log`

This endpoint and the already existing

* GET `/api/sources/feeds/:id/log`

now have three more query parameters besides the already existing filtering by log levels:

* `from` start timestamp from which the log will be exported.
* `to` timestamp to which the log will be exported.
* `search` a search string looked up in the message of the logs.

This PR also adds a field `feed_id` to the output log entries.

### DIFF
--- a/pkg/database/query/sqlbuilder.go
+++ b/pkg/database/query/sqlbuilder.go
@@ -40,7 +40,9 @@ var (
 	whiteSpaces = regexp.MustCompile(`\s+`)
 )
 
-func likeQuery(query string) string {
+// LikeEscape quotes a query string to be more convenient
+// to use with LIKE filters.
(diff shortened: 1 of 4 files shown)


koplas (approved):
LGTM & works

PR #639, opened 2024-11-14
Avoid ad block preventing the download of logs

koplas:
The popular EasyPrivacy list blocks all requests that start with `/log?count=`. Adding a null op parameter to the request avoids this issue.


s-l-teichmann:
Not an ideal solution, because adblockers may adjust to parameter permutations, but okay for the moment. Retested it and it works ... for now.

s-l-teichmann (approved):
L(not so)GTM & works

Issue #596, opened 2024-11-09
Error on Dashboard on PostgreSQL 15


tschmidtb51:
As `beate` I see an error on the dashboard:

![grafik](https://github.com/user-attachments/assets/5605ba90-9961-4f8f-951a-7b3dfe8c8b9b)

I run all migrations after the last build.

ISDuBA: 0.4.1-1-g75fe766f

bernhardreiter:
@tschmidtb51 can you check your postgresl version.

Either as SQL:
```sql
SELECT version();
```

or like `pg_config --version`

tschmidtb51:
Looks like I have a PostgreSQL 17.0 (shown via `pg_config --version`) and a PostgreSQL 15.8 (via `SELECT version();`). According to `\l` when logged into the database, I'm currently using 15.8 (as there are the databases). Do you have any suggestions how to migrate from 15.8 to 17.0?

s-l-teichmann:
We've add a path with PR #644 which may help here. We cannot finally confirm it yet as we currently don't have a PG15 system at hand.

tschmidtb51:
> We've add a path with PR #644 which may help here. We cannot finally confirm it yet as we currently don't have a PG15 system at hand.

I just pulled and build it and the error is gone on PostgreSQL 15. So I believe it was fixed.

PR #644, opened 2024-11-14
Add alias to event comment subquery

koplas:
This avoids issues with PostgreSQL versions older than 16, that don't support subqueries without alias.


s-l-teichmann:
I can confirm that it works with PG16. In the hope that it will work with PG15 I'm  going to merge it.
The real prove still has to be done as I've currently no PG15 setup. So we keep #596 open till then.

PR #645, opened 2024-11-14
feat: Give advisory version buttons dynamic width

luisa-beerboom:
Turned former fixed width into min width to ensure the button gets wider on longer version numbers

This ensures that long version numbers don't flow out of the button like in this example:
Screenshot_20241114_160354


cintek (approved):
LGTM & works

PR #654, opened 2024-11-15
Improve toplevel README: move to v0.5.0

bernhardreiter:
* Switch from pre-release to beta quality. For this move status lower.
  * Add hint to invite reports.
  * Remove a few typos in other parts.
  * Move go lower in the list of used components as the database has a higher importance for potential users now that we are in beta.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,21 +10,24 @@
 
 # ISDuBA
 
-**Under development -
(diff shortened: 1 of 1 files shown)


tschmidtb51 (approved):
LGTM

PR #656, opened 2024-11-15
fix: build-in version for release tags

### LINKED ISSUE #651: ISDuBA Version
A version should not have the leading `v` according to SemVer. For Git tags, it totally makes sense to have it (and many projects do). Maybe, we should just remove it before the build :thinking: 

@bernhardreiter / @s-l-teichmann: Thoughts?

bernhardreiter:
* Change Makefile to remove the leading `v` from the git tag in the case of release tags. Previously this was only done for pre-release git tags.

resolve #651


koplas (approved):
LGTM & works

PR #660, opened 2024-11-17
Simplify source checking

s-l-teichmann:
This PR simplifies the check if source checking should be prevented when already running.
The original code had a much more complicated state management in  mind which turns out not to be necessary.


koplas (approved):
LGTM & works

PR #666, opened 2024-11-19
Add default aggregator configuration

### LINKED ISSUE #655: Add default CSAF aggregator of the BSI, allow configuration of the default aggregator

koplas:
Closes #655


s-l-teichmann (commented):
No for many reasons:
* Why have a config if we also have it in the database?
* In the migration we should check ig we already have it.
* See issue about being active by default.

In short: We have to talk about this.

koplas:
Superseded by #682

Issue #667, opened 2024-11-19
Docker compose setup issues


manucert:
I have two different issues using ISDuBA with docker.

Using the Windows WSL Docker setup I get following issue during the startup (docker compose up):
`isduba-keycloak        | 2024-11-19 14:43:00,619 ERROR [org.keycloak.quarkus.runtime.configuration.mappers.PropertyMappers] (main) Hostname v1 options [hostname-url] are still in use, please review your configuration
isduba-keycloak        | 2024-11-19 14:43:06,287 INFO  [org.keycloak.quarkus.runtime.storage.infinispan.CacheManagerFactory] (main) Starting Infinispan embedded cache manager
isduba-keycloak        | 2024-11-19 14:43:06,679 INFO  [org.keycloak.quarkus.runtime.storage.infinispan.CacheManagerFactory] (main) Persistent user sessions enabled and no memory limit found in configuration. Setting max entries for sessions to 10000 entries.
isduba-keycloak        | 2024-11-19 14:43:06,680 INFO  [org.keycloak.quarkus.runtime.storage.infinispan.CacheManagerFactory] (main) Persistent user sessions enabled and no memory limit found in configuration. Setting max entries for clientSessions to 10000 entries.
isduba-keycloak        | 2024-11-19 14:43:06,680 INFO  [org.keycloak.quarkus.runtime.storage.infinispan.CacheManagerFactory] (main) Persistent user sessions enabled and no memory limit found in configuration. Setting max entries for offlineSessions to 10000 entries.
isduba-keycloak        | 2024-11-19 14:43:06,680 INFO  [org.keycloak.quarkus.runtime.storage.infinispan.CacheManagerFactory] (main) Persistent user sessions enabled and no memory limit found in configuration. Setting max entries for offlineClientSessions to 10000 entries.
isduba-db              | 2024-11-19 14:43:07.365 UTC [109] FATAL:  database "keycloak" does not exist`

I used the default .env, therefore keycloak and isduba is set to listen on localhost, but the possible reason for the issue above is probably earlier. I guess it is also something Windows Docker Plattform specific, because it seems empty rows are an issue there:
` /docker-entrypoint-initdb.d/init.sh: line 8: $'\r': command not found
: invalid option       | /docker-entrypoint-initdb.d/init.sh: line 10: set: -
isduba-db              | set: usage: set [-abefhkmnptuvxBCEHPT] [-o option-name] [--] [-] [arg ...]
[...]

PR #669, opened 2024-11-20
Migrate Keycloak config and cleanup

koplas:
Relevant #667

### DIFF
--- a/docker/server/Dockerfile
+++ b/docker/server/Dockerfile
@@ -19,7 +19,7 @@ COPY Makefile Makefile
 RUN make build_client
 
 
-FROM library/golang:1.22.1-alpine AS server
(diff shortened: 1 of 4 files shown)


docker/server/Dockerfile
```diff
@@ -19,7 +19,7 @@ COPY Makefile Makefile
 RUN make build_client
 
 
-FROM library/golang:1.22.1-alpine as server
+FROM library/golang:1.22.1-alpine AS server
```
s-l-teichmann, line 22: Side remark: 1.22.1 is pretty old. Is there an image for say 1.22.9?

koplas: Updated to the latest stable.

cintek (approved):
LGTM & works

PR #671, opened 2024-11-20
Save filename on import

### LINKED ISSUE #668: safe filename from bulk import
We need to safe the filename when using the bulk importer.
(As filename persistence was introduced later.)

koplas:
The filename was only saved if the source manager downloaded it. Now the bulk importer and the import API also save the filename.

Closes #668


s-l-teichmann (approved):
LGTM & works

PR #672, opened 2024-11-20
fix: Show stats of manual imports in source overview

cintek:
IIRC we had them already but probably they were removed by accident when domain etc. from the fake source were removed.

import_stats


koplas (approved):
LGTM & works

PR #673, opened 2024-11-20
Serve documents with their filenames

s-l-teichmann:
This PR sets the Content-Disposition header to the filename of a document
to the stored one from the database.

When using the bulkimporter now ".gz" extensions are stripped from the filename.


koplas (approved):
LGTM & works

PR #676, opened 2024-11-21
Use the correct name of the standard

cintek:
OpenPGP is the name of the standard while PGP is the name of an encryption program.


koplas (approved):
LGTM

PR #677, opened 2024-11-21
Improvement of stats

cintek:
Show loading indicator next to header -> prevent jumping when selecting date range
stats_loading

Also improve wording of a label used in a legend (see 0dc309d36d0b4d27c91836afd2a28bc07f9ed86d).


koplas (approved):
LGTM

PR #678, opened 2024-11-21
Aggregator list part 1

cintek:
This is the first part of changes of #665.


koplas (approved):
LGTM & works

cintek:
Since we finally finished #665 we don't need this anymore.

PR #683, opened 2024-11-21
Restrict HTTP dialing to remote hosts

### LINKED ISSUE #680: Add ip connection blacklist for localhost and internal networks
By design isdubad (on the server) will follow URLs determined partly by external contents, see #679.

As precaution and a secure default configuration we want to  add an ipv4 and ipv6 level blacklist and preconfigure it with `localhost` ips and internal networks like `168.`, `10.` or `192.` .

If someone then wants to make use of an internal "provider", this configuration must allow to open one or several ip addresses again.

s-l-teichmann:
Resolves #680 

This PR add support to limit the HTTP(S) access of the source manager (handling PMDs, feed indices and advisory downloads) and aggregators.

As these components are intended to load content from remote hosts there is a risk to use these
as e.g. service scanning devices in form of blind Server-Side Request Forgery (SSRF).

To minimize this effect four new configuration parameters are introduced.

1. `allowed_ports` is a list of ports that are allowed to be contacted. This defaults to the port `80` and `443`.
2.  `block_loopback` enables blocking access to loopback devices. This defaults to `true`.
3. `blocked_ranges` is a list of IP sub nets to which access will be blocked. The default protects typical local sub nets:
     -  `127.0.0.0/8`        IPv4 loopback                                    
     -  `10.0.0.0/8`          RFC1918                                          
     -  `172.16.0.0/12`    RFC1918                                          
     - `192.168.0.0/16`   RFC1918                                          
     -  `169.254.0.0/16`  RFC3927 link-local                               
     - `::1/128`                 IPv6 loopback                                    
     - `fe80::/10`              IPv6 link-local                                  
     - `fc00::/7`                 IPv6 unique local addr
4. `allowed_ips`. This is list is checked if an IP is blocked by an entry of `blocked_ranges` to punch in specific holes. By default this list is empty.

### DIFF
--- a/cmd/isdubad/main.go
+++ b/cmd/isdubad/main.go
@@ -61,7 +61,7 @@ func run(cfg *config.Config) error {
 	forwardManager := forwarder.NewForwardManager(&cfg.Forwarder, db)
 	go forwardManager.Run(ctx)
 
-	agg := aggregators.NewManager(&cfg.Aggregators, db)
+	agg := aggregators.NewManager(cfg, db)
 	go agg.Run(ctx)
 
 	// Is the remote validator configured?
--- a/docs/example_isdubad.toml
+++ b/docs/example_isdubad.toml
@@ -9,6 +9,18 @@
 # [general]
 # advisory_upload_limit = "512K"
 # anonymous_event_logging = false
+# allowed_ports = [80, 443]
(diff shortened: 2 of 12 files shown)


koplas (approved):
LGTM & works

PR #687, opened 2024-11-25
Remove attention flag if feeds are saved

### LINKED ISSUE #681: Remove attention flag when created a new source
In the process of creating a new source the user can already see which feeds will be configured. That's why we should remove the attention flag right away by sending a request from the client when the source was created.

koplas:
Closes #681


cintek (approved):
LGTM & works

PR #689, opened 2024-11-26
Add remark about server side request forgery

### LINKED ISSUE #679: Docs: add recommendation to place isdubad outside of internal networks
Because the system is designed to download CSAF document from all over the internet, the backend isdubad can actually make a lot of request.

A `source-manager` or someone controlling PMDs or aggregator.json files, can insert URLs that isdubad would call.

Thus a running instance of ISDuBA should have the server placed in a network segment where no internal networks are accessibile and the process should also be limited to not be able so access other port on localhost.

We need to document this operating requirement.

s-l-teichmann:
Resolves #679

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,7 @@ ISDuBA uses:
 ## How to get started
 
 - [Setup](docs/setup.md)
+- [Operations](docs/operations.md)
 
 ## What does the name ISDuBA mean?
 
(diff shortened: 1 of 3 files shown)


bernhardreiter (approved):
Added more details and changes `URLs` -> `IP addresses`.

Now: LGTM

PR #690, opened 2024-11-26
Small adjustments in the aggregator list

cintek:
1. Don't use orange color for rounded boxes.
2. Remove indention of sources.
3. First list configured then unconfigured feeds.
4. Remove small source button and make the whole source to a button.

aggr


JanHoefelmeyer (commented):
Can the visibility while hovering be increased in light mode as well?
See:
Screenshot from 2024-11-27 12-41-04

PR #693, opened 2024-11-26
Statistics: Delay requests when user selects dates and abort old requests

### LINKED ISSUE #688: Statistics time range can send way too many requests
Whenever the time range for statistics is updated, a new request is send to the server.

Using, for example, the year-setter and scrolling up and down with the mouse-wheel will send a lot of requests in quick succession, which has caused my server to become completely unresponsive indefinitely.

cintek:
Otherwise, when the user uses the scroll wheel to choose the month/year of a date and is not scrolling fast enough one request per month/year is made.

With this commit requests also are aborted if new ones are made. In the end these changes lead to less load for the backend.

This PR affects only the selection of `from` and `to` and fixes #688 . Notice the new dependency. For testing it is required to run `npm install`.

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -14,6 +14,7 @@
         "chart.js": "^4.4.6",
         "chartjs-adapter-moment": "^1.0.1",
         "chroma-js": "^3.1.2",
+        "debounce": "^2.2.0",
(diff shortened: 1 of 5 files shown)


JanHoefelmeyer (approved):
LGTM & works

PR #694, opened 2024-11-26
Sources: Filters for feed logs

cintek:
PRs #349 and #635 allow to search and filter feed logs. In this PR this functionality will be offered to the user in the UI.

Filter by:
- [ ] Log level
- [ ] Time range
- [ ] Search term


cintek:
In the end I decided to keep the pagination but add the count to it. So it also shows if there are no entries. But the table will be removed when there are no entries.
pagination

JanHoefelmeyer (approved):
LGTM & works

JanHoefelmeyer (dismissed):
LGTM & works

PR #698, opened 2024-11-26
Cleaning feed logs

### LINKED ISSUE #697: Time limit the storage of download logs
Currently the download logs grow indefinitely which will fill up the database on the long run.

It would be a better idea to keep the logs only a certain amount of time like 3 month
and delete the log entries afterwards. The time limit should be configurable and
should even able to be turned off.

s-l-teichmann:
Resolves #697

This PR adds  a configuration variable `keep_feed_logs` defaulting to ~3 months (3*31 days).
Feed log entries older than this value will be removed from the database every 20minutes.
Setting this value to zero or less will deactivate the removal.

### DIFF
--- a/docs/example_isdubad.toml
+++ b/docs/example_isdubad.toml
@@ -83,6 +83,7 @@
 # timeout = "30s"
 # default_age = "17520h"
 # checking = "2h"
+# keep_feed_logs = "2232h"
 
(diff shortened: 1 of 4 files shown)


JanHoefelmeyer (approved):
LGTM & works

Issue #504, opened 2024-10-15
Search-Event-Tab: Restrict to users which can fetch events


JanHoefelmeyer:
Currently, Admin, Source-Manager and Importer will get a 403 when visiting the Event tab of the search page.

Default-Queries might need adjustments too.

PR #700, opened 2024-11-27
Fix: Restrict roles that cannot request events from viewing the events tab

JanHoefelmeyer:
closes https://github.com/ISDuBA/ISDuBA/issues/504

Admin-role is planned to be able to view events, but currently can't. 

Commit d3fa000047efa090ef0abf848c079d85e8fcda25 enables admins to view events, even if it causes an error for now. -> https://github.com/ISDuBA/ISDuBA/pull/702 solves this error

### DIFF
--- a/client/src/lib/Search/Overview.svelte
+++ b/client/src/lib/Search/Overview.svelte
@@ -201,17 +201,19 @@
           clearSearch();
         }}>Documents</Button
       >
-      <Button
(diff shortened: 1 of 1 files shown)


s-l-teichmann (approved):
LGTM

cintek (approved):
LGTM & works

PR #702, opened 2024-11-27
Allow admin to view and edit comments, give access to events, too.

### LINKED ISSUE #554: Rework Events permissions to fit with current usecases
There are some places where missing permissions to see events cause issues. These can be mitigated by having also a role that can see events, but will lead to errors if not.

This affects things like the "X-recent-global-default"-Query on the Dashboard and the Events-Tab on the Search page.

Also, admins cannot see or respond to pings which one might want to do to indicate why an advisory should be deleted or whether it should be reverted from deletion status if put there erroneously.

s-l-teichmann:
Resolves #554 

This PR adds permissions for admin to work work with comments end events.

### DIFF
--- a/client/src/lib/Advisories/Advisory.svelte
+++ b/client/src/lib/Advisories/Advisory.svelte
@@ -50,22 +50,26 @@
   let processRunning = false;
   let lastSuccessfulForwardTarget: number | undefined;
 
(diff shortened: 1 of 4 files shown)


cintek (approved):
LGTM & works

PR #703, opened 2024-11-27
Feat: Reenable recent changes query for admin on dashboard

JanHoefelmeyer:
Due to https://github.com/ISDuBA/ISDuBA/pull/702, admins can now properly use the recent changes query too so there is no longer  a need to hide them. (Requires https://github.com/ISDuBA/ISDuBA/pull/702)


cintek (approved):
LGTM & works

PR #708, opened 2024-11-28
Feed logs: Add time filter

cintek:
Same as #694 but with time inputs so the user can filter the logs by more specific time frames.


JanHoefelmeyer (approved):
LGTM & works

Issue #670, opened 2024-11-20
Use green color for save buttons


cintek:
Instead of white use green (-> positive meaning) color for save buttons.

PR #710, opened 2024-11-28
Change colors of buttons

cintek:
* Green for save buttons
* Don't highlight the "cancel" button in the SSVC calculator with primary color because you usually highlight buttons which lead the user forward to their goal

Solves #670


JanHoefelmeyer (approved):
Buttons are consistently green as far as I can tell

LGTM & works

PR #713, opened 2024-11-29
Use PostgreSQL 15 for testing and development

koplas:
This allows the detection of the usage of new features that are not available in PostgreSQL 15.

### DIFF
--- a/docs/scripts/installpostgres.sh
+++ b/docs/scripts/installpostgres.sh
@@ -10,5 +10,10 @@
 
 set -e # to exit if a command in the script fails
 
-# install PostgreSQL 16
-sudo apt install postgresql -y
+# install PostgreSQL 15
+sudo apt install ca-certificates -y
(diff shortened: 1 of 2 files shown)


docs/scripts/installpostgres.sh
```diff
@@ -10,5 +10,10 @@
 
 set -e # to exit if a command in the script fails
 
-# install PostgreSQL 16
-sudo apt install postgresql -y
+# install PostgreSQL 15
+sudo apt install wget ca-certificates -y
```
bernhardreiter, line 14: As we have curl already, please replace wget with curl to avoid another unnecessary dependency.

bernhardreiter (changes requested):
Needs wget -> curl simplification

docs/scripts/installpostgres.sh
koplas: Done

PR #714, opened 2024-11-29
Expose keep_feed_logs configuration to client

koplas:


s-l-teichmann (changes requested):
It seems a bit unmotivated here.

koplas:
This was based on a request from @cintek. @cintek can you explain the motivation?

cintek:
@JanHoefelmeyer suggested using the value of `keep_feed_logs` to pre-fill the `from` date where the feed logs are so we don't have date inputs which are just showing the placeholder.

JanHoefelmeyer:
This shouldn't be done without a new explaining comment via the defaultMessage function but in a proper place with proper explanation

PR #724, opened 2024-12-02
feat: Regular updates in sources with loading bars

### LINKED ISSUE #573: Source Management: Import stats not updated
In `Sources`  there are these overviews:

Source overview:
source_stats3

Overview of single source:
source_stats

While the number of "Loading" and "Queued" are updated in small intervals those of "Imported" are not. To be consistent they also should be updated in the same interval.

luisa-beerboom:
'Imported' statistic updated every 30 seconds, everything else every 5 seconds

Closes #573

### DIFF
--- a/client/src/app.pcss
+++ b/client/src/app.pcss
@@ -49,3 +49,37 @@ Webview styles
   animation-duration: 4s;
   border-radius: 0.5rem;
 }
+
+.progressmeter {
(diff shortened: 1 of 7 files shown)


JanHoefelmeyer (approved):
Would be nice to see how far the charge bars go for charging. But LGTM & works

PR #728, opened 2024-12-05
Unix domain sockets

s-l-teichmann:
This PR introduces the possible use of unix domain socket to connect to the database and serving isduabd behind a reverse proxy like caddy which can proxy tcp to domain socket.

This only the first step. The documentation how to use this in detail has to be improved.


koplas (approved):
LGTM & works

PR #730, opened 2024-12-05
Markdown autodocs

cintek:
- don't mention `git` explicit because developers can look into the scripts for the details
- update usage of install script
- adjust parts of the scripts so it can be used in the docs via autodocs


JanHoefelmeyer (approved):
LGTM

PR #732, opened 2024-12-05
feat: Temporarily show button for selected dashboard query

### LINKED ISSUE #720: Use an Indicator when using a Dashboard query on the Search tab
On the desktop, you are shown 6 advisories via a predefined query. Seeing more advisories of that query requires you to press a button below the query-column that opens the query on the search page. However, there's no indicator on the search page that this happened and that you're seeing a filtered and thus possibly reduced amount of advisories. Such an indicator should be added.

luisa-beerboom:
Closes #720

### DIFF
--- a/client/src/lib/Search/Overview.svelte
+++ b/client/src/lib/Search/Overview.svelte
@@ -47,13 +47,13 @@
     advisoryTable.fetchData();
   };
 
-  $: if (selectedCustomQuery === -1) {
+  $: if (selectedCustomQuery < 0) {
     setQueryBack();
   }
 
   const triggerSearch = async () => {
     if (!advancedSearch) {
-      if (selectedCustomQuery === -1) {
+      if (selectedCustomQuery < 0) {
         query.query = searchTerm ? `"${searchTerm}" search ${searchColumnName} as` : "";
       } else {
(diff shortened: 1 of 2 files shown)


cintek (changes requested):
It is working that the opened queries are visible in Search but there is one problem:

- Login as editor
- click on the button below the **event** queries
more-button
- select "Documents"
opened-dashboard-query

Result:
opened-dashboard-query-error

luisa-beerboom:
@cintek can't replicate this. Do you have an up-to-date backend?

luisa-beerboom:
The client is calling with the route `api/documents?query=$state new workflow =&advisories=false&count=1&orders=-critical&limit=10&offset=0&columns=critical cvss_v3_score cvss_v2_score ssvc four_cves publisher title tracking_id initial_release_date current_release_date version comments id`
IDK why that causes an error, I'm not really knowledgable about the query request syntax.
@s-l-teichmann What do you think?

JanHoefelmeyer (approved):
LGTM
The issue described by @cintek is indeed in main as well and now tracked by
https://github.com/ISDuBA/ISDuBA/issues/752

cintek (approved):
Works

Issue #711, opened 2024-11-28
Advisory view: Block user from entering invalid ssvc


JanHoefelmeyer:
Users can currently enter any string as ssvc and attempt to safe it, leading to an error message. We should probably only allow trying to save on valid ssvc

PR #736, opened 2024-12-05
feat: Add SSVC validation in client

cintek:
Solves #711.

### DIFF
--- a/client/src/lib/Advisories/SSVC/SSVCCalculator.svelte
+++ b/client/src/lib/Advisories/SSVC/SSVCCalculator.svelte
@@ -9,11 +9,12 @@
 -->
 
 <script lang="ts">
-  import { Button, Label, Input, StepIndicator } from "flowbite-svelte";
+  import { Button, Label, StepIndicator } from "flowbite-svelte";
   import {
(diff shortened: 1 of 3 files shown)


JanHoefelmeyer (changes requested):
Opening the SSVC dialogue:
Uncaught (in promise) TypeError: date is undefined
    isDateValid SSVCInput.svelte:50
    validateVector SSVCInput.svelte:44
    instance SSVCInput.svelte:31
    run utils.js:41
    mount_component Component.js:47
    flush scheduler.js:99

cintek:
Current state: When @JanHoefelmeyer is testing the PR and he pastes a SSVC into the input it isn't validated right away. Only if he moves the cursor within the field or enters a character. I cannot reproduce the issue. What is expected is that the validation starts even if a value was only pasted and no further interaction was made and when I test this it works as expected.

cintek:
The problem was that @JanHoefelmeyer used the mouse button to paste the value but it wasn't recognized as expected. This should be fixed now.

JanHoefelmeyer (approved):
LGTM & works

PR #737, opened 2024-12-05
fix: Make QueryTable name and description columns use text-wrap

luisa-beerboom:
Previously, setting long titles or descriptions for queries would cause them to be displayed inline and cause the columns in question to expand to the width of the longest entry. The table could therefore flow out of the page, even with bigger screen sizes.

This commit causes long names and descriptions to wrap if they get too long.


cintek (changes requested):
It works if the strings contain white spaces but if the user enters long strings without the text is still overflowing.

cintek (approved):
LGTM & works

Issue #741, opened 2024-12-06
Show loading indicator when downloading feed logs


cintek:

koplas:
Closed with #742.

PR #742, opened 2024-12-06
Improve UX of downloading feed logs

cintek:
* show indicator and explanation while download is prepared
* make downloads cancelable

Solves #741


koplas (approved):
LGTM & works

Issue #725, opened 2024-12-03
Allow the backend to connect to postgresql via unix domain sockets (for single machine performance)


bernhardreiter:
Tests suggest that using unix domain sockets for the PostgreSQL database can gain significant performance on machines where everything is setup in the same GNU/Linux installation. (As opposed to multi-VM or multi-machine setups.)

s-l-teichmann:
Plus using Domain Sockets for isdubad itself If IT Runs behind a Proxy Like Caddy which allows proxying TCP to Domain Sockets.Am 03.12.2024 14:27 schrieb "Bernhard E. Reiter" ***@***.***>:
Assigned #725 to @s-l-teichmann.

s-l-teichmann:
PR #728 implements the basics.

PR #743, opened 2024-12-06
Cleanup domain socket at shutdown

s-l-teichmann:
This PR shuts down the isdubad server properly if a KILLTERM signal is send from systemd.
It furthermore removes the created socket file at shutdown.
The port is interpolated into the host path in case of the use of domain sockets.

Contributes to #725


koplas (approved):
LGTM & works

PR #747, opened 2024-12-10
Don't listen to all network interfaces in PostgreSQL

koplas:
This avoids exposing ports that are not required.


JanHoefelmeyer (approved):
works

PR #749, opened 2024-12-10
Add diagram as overview of ISDuBA's components

cintek:


s-l-teichmann (dismissed):
Please explain to me why are the "Download" boxes not next to the "Sources"? In fact we download from the sources.

cintek:
> Please explain to me why are the "Download" boxes not next to the "Sources"? In fact we download from the sources.

Good objection. I removed the download boxes completely with the latest change.

s-l-teichmann (approved):
LGTM

PR #753, opened 2024-12-11
Feat: Add important docs into the dist tarball

JanHoefelmeyer:
When building the docs (not the scripts etc.) now are also included into the tar archive


koplas (approved):
LGTM & works

PR #755, opened 2024-12-11
Fix comment edit

koplas:
If a comment is written and then edited, an older comment message was shown. This fixes the reactivity issue.


JanHoefelmeyer (approved):
LGTM & works

PR #756, opened 2024-12-11
Add ca-certificates to docker setup

### LINKED ISSUE #754: TLS error when adding sources - docker compose setup
I tried adding the BSI aggregator by inputting the aggregator.json URL (below). But I get following error after adding the new aggregator:
`Could not fetch aggregator data 400: The request sent could not be understood. Get "https://wid.cert-bund.de/.well-known/csaf-aggregator/aggregator.json": tls: failed to verify certificate: x509: certificate signed by unknown authority`

The same when adding as a CSAF trusted provider.

Is this possibly because of the docker setup or why do I get a certificate error? 
(I used the default docker setup without any modification - just compose build and up)

koplas:
This is required to download documents from secure CSAF trusted providers and CSAF aggregators.

Closes #754

### DIFF
--- a/docker/server/Dockerfile
+++ b/docker/server/Dockerfile
@@ -21,6 +21,8 @@ RUN make build_client
 
 FROM library/golang:1.23-alpine AS server
 RUN apk update && apk add --no-cache make
+# This required to download from https locations, like CSAF trusted provider or CSAF aggregator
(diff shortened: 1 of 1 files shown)


cintek (approved):
LGTM & works

PR #758, opened 2024-12-11
Split white-green group into white and green

koplas:
This is more clean than defining a group that contains both TLP.

### DIFF
--- a/docker/keycloak/init.sh
+++ b/docker/keycloak/init.sh
@@ -61,10 +61,15 @@ kcadm.sh update clients/$id --target-realm=isduba \
     } } ]'
 
 # Create groups
-WHITE_GREEN_ID=$(kcadm.sh create groups --target-realm=isduba --set name=white-green -i)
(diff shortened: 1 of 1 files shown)


docker/keycloak/init.sh
```diff
@@ -61,10 +61,15 @@ kcadm.sh update clients/$id --target-realm=isduba \
     } } ]'
 
 # Create groups
-WHITE_GREEN_ID=$(kcadm.sh create groups --target-realm=isduba --set name=white-green -i)
+WHITE_ID=$(kcadm.sh create groups --target-realm=isduba --set name=white -i)
 kcadm.sh update groups/$WHITE_GREEN_ID --target-realm isduba \
```
JanHoefelmeyer, line 65: Does WHITE_GREEN_ID even exist here anymore?

koplas: Done

JanHoefelmeyer (approved):
LGTM

PR #760, opened 2024-12-11
Replace pg_input_is_valid with PG15 compatible expression.

s-l-teichmann:
pg_input_is_valid is PG16. We target PG15.
So use a regular expression to check if version is an integer.


koplas (approved):
LGTM & works (in PG16)

PR #763, opened 2024-12-12
Allow to user other filenames for the config

cintek:
Since not only the BSI will be using this application we should allow other filenames for the config without users having to fear to commit the config by accident.


JanHoefelmeyer (approved):
LGTM & works

PR #767, opened 2024-12-12
Fix query selection detection

### LINKED ISSUE #752: Error when switching query type when using the "more"-button on the dashboard
The more buttons basically execute a query, yet don't hide the type buttons allowing switching to types the query was not made for (e.g. events on a document query). This causes an error.

koplas:
If a query from the dashboard was selected, the type selection button was not hidden. This fixes it by correctly setting the state if the query was specified via URL parameter.

Closes #752


JanHoefelmeyer (approved):
works

PR #773, opened 2024-12-16
Fix likepid documentation

koplas:
`ilikepid` only searches for product IDs and not product names.

### DIFF
--- a/docs/filter_expr.md
+++ b/docs/filter_expr.md
@@ -30,7 +30,7 @@ The data type of the `$cvss_v3_score` is `float`.
 
 To compare another value to this value the other value has to be of a compatible data type.
 Writing simply `5` would be treated as a `string`.  
-To make it a `float` we have to **cast** it by appending the required type to it: `5 float`.  
+To make it a `float` we have to **cast** it by appending the required type to it: `5 float`.
 
 Now `>=` as an **operator** can be applied to `$cvss_v3_score` as
 both operands are of type `float`. `>=` is a condition evaluating to `true` or `false`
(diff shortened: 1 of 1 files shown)


docs/filter_expr.md
```diff
@@ -45,124 +45,123 @@ $cvss_v3_score 5 float >= $current_release_date 2023-12-31 timestamp > and
...
+| `=`         | **A** **B**           | `bool` **A** equals **B**                                                                                 |
+| `!=`        | **A** **B**           | `bool` **A** not equals **B**                                                                             |
+| `<`         | **A** **B**           | `bool` **A** lesser than **B\***                                                                          |
+| `<=`        | **A** **B**           | `bool` **A** lesser or equal than **B**                                                                   |
+| `>`         | **A** **B**           | `bool` **A** greater than **B**                                                                           |
+| `>=`        | **A** **B**           | `bool` **A** greater or equal than **B**                                                                  |
+| `ilike`     | `string` `string`     | `bool` First argument is case insensitive like second argument                                            |
+| `ilikepid`  | `string`              | `bool` Is there a product in the product tree with a name like the argument?                              |
```
bernhardreiter, line 115: This seems to be the same as the original, so the problem is not fixed.
(Please do not make unrelated whitespace changes in a pull request. It makes reviews much harder.)

bernhardreiter (dismissed):
not fixed (as far as I can see through the thicket of whitespace. ;) )

cintek (approved):
Usually, it would be better to separate whitespace fixes into another commit so it's easier to recognize the essential changes. But otherwise LGTM.

PR #785, opened 2025-01-05
OpenAPI support

s-l-teichmann:


pkg/web/sources.go
```diff
@@ -563,23 +661,34 @@ func (c *Controller) updateFeed(ctx *gin.Context) {
...
 	default:
 		slog.Error("database error", "err", err)
-		ctx.JSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
+		models.SendError(ctx, http.StatusInternalServerError, err)
 	}
 }
 
+// viewFeed is an endpoint that returns all feeds.
+//
+//	@Summary		Returns feeds.
```
cintek, line 677: Only a single feed

pkg/web/sources.go
```diff
@@ -588,31 +697,52 @@ func (c *Controller) viewFeed(ctx *gin.Context) {
...
+		models.SendErrorMessage(ctx, http.StatusNotFound, "feed not found")
 		return
 	}
 	ctx.JSON(http.StatusOK, newFeed(fi))
 }
 
+// deleteFeed is an endpoint that deletes the feed with specified id.
+//
+//	@Summary		Deletes a feed.
+//	@Description	Deletes the feed configuration with the specified id.
```
cintek, line 709: Should be "ID"

pkg/web/sources.go
```diff
@@ -706,27 +848,50 @@ func (c *Controller) feedLogs(ctx *gin.Context, feedID *int64) {
...
-		h["count"] = counter
+		h.Count = &counter
 	}
 	ctx.JSON(http.StatusOK, h)
 }
 
 // defaultMessage returns the default message.
+//
+//	@Summary		Returns the default message.
+//	@Description	Returns the message that is displayed on visiting the sources page.
```
cintek, line 864: IMHO this is a bit too short. It should be explained that this message is shown to users to explain that they can contact the source manager to add new or edit sources.

pkg/web/sources.go
```diff
@@ -220,34 +245,56 @@ func (c *Controller) createSource(ctx *gin.Context) {
...
 	default:
 		slog.Error("database error", "err", err)
-		ctx.JSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
+		models.SendError(ctx, http.StatusInternalServerError, err)
 	}
 }
 
+// viewSource is an endpoint that returns information about the source.
+//
+//	@Summary		Get source information.
```
cintek, line 289: Should be "Returns source information."

pkg/web/sources.go
```diff
@@ -220,34 +245,56 @@ func (c *Controller) createSource(ctx *gin.Context) {
...
 		slog.Error("database error", "err", err)
-		ctx.JSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
+		models.SendError(ctx, http.StatusInternalServerError, err)
 	}
 }
 
+// deleteSource is an endpoint that deletes the source with specified id.
+//
+//	@Summary		Deletes a source.
+//	@Description	Deletes the source configuration with the specified id.
```
cintek, line 260: Should be "ID"

pkg/web/sources.go
```diff
@@ -466,6 +524,22 @@ func validateHeaders(headers []string) error {
...
 
+type feedResult struct {
+	Feeds []*feed `json:"feeds"`
+}
+
+// viewFeeds is an endpoint that returns all feeds.
+//
+//	@Summary		Returns feeds.
+//	@Description	Returns all feed configurations and metadata.
+//	@Param			id		path	int		true	"Feed ID"
```
cintek, line 535: This has to be "Source ID"

pkg/web/advisory.go
```diff
@@ -110,53 +110,92 @@ func (c *Controller) changeStatusAll(ctx *gin.Context, inputs advisoryStates) {
...
+		models.SendErrorMessage(ctx, http.StatusBadRequest, "state transition not possible")
 	default:
-		ctx.JSON(http.StatusOK, gin.H{"message": "transition done"})
+		models.SendSuccess(ctx, http.StatusOK, "transition done")
 	}
 }
 
+// changeStatus changes the status the specified advisories.
+//
+//	@Summary		Changes the status of the advisory.
```
cintek, line 134: I would use "of **an** advisory"

pkg/web/tempdocuments.go
```diff
@@ -55,24 +66,46 @@ func (c *Controller) importTempDocument(ctx *gin.Context) {
...
-		"free":  free,
+	ctx.JSON(http.StatusOK, tempDocuments{
+		Files: files,
+		Free:  free,
 	})
 }
 
+// viewTempDocument is an endpoint that returns a temporary document with the specified id.
+//
+//	@Summary		Returns a temporary documents.
```
cintek, line 100: Singular: "document"

pkg/web/tempdocuments.go
```diff
@@ -100,15 +133,25 @@ func (c *Controller) viewTempDocument(ctx *gin.Context) {
 		extraHeaders)
 }
 
+// deleteTempDocument is an endpoint that deletes a temporary document with the specified id.
+//
+//	@Summary		Deletes a temporary documents.
```
cintek, line 138: Singular: "document"

pkg/web/tempdocuments.go
```diff
@@ -100,15 +133,25 @@ func (c *Controller) viewTempDocument(ctx *gin.Context) {
...
+//	@Description	Deletes a temporary document with the specified id.
```
cintek, line 139: Should be "ID"

koplas:
A list of endpoints where the input form is currently missing:
- `/status`
- `/sources/:id` (PUT)
- `/sources/feeds/:id/log` (PUT)
- `/sources/feeds/:id` (PUT)
- `/documents` (POST)

pkg/web/sources.go
koplas: The default message can be configured to display anything. If we document this, we should enforce that the default message cannot be changed.

pkg/web/events.go
```diff
@@ -25,6 +25,17 @@ import (
 	"github.com/ISDuBA/ISDuBA/pkg/models"
 )
 
+// overviewEvents is an endpoint that returns a list of events.
+//
+//	@Summary		Returns a list of events.
+//	@Description	Returns all events that match the specified query.
```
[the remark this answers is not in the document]
koplas: You can specify something via the query parameter.

pkg/web/events.go
```diff
@@ -25,6 +25,17 @@ import (
...
+//	@Param			query	query	string	false	"Event query"
```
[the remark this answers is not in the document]
koplas: See above.

JanHoefelmeyer (changes requested):
Just about every request results in a 401 if unauthorized (which can happen on the swagger site or if using ISDuBA via command line), this needs to be documented as well

koplas:
> Besides the things I commented in the code I have to general things: For some reason when you have durations in the examples a negative value is shown: time This might be confusing.
> 
> Then there are endpoints where you have to pass form data but this is not visible in the endpoints: body

This will be fixed in later iterations.

koplas (approved):

JanHoefelmeyer (approved):

JanHoefelmeyer (approved):
LGTM & works

PR #792, opened 2025-01-07
Remove progress bar for `Loading/Queued`

koplas:
This progress bar moves too fast, which may irritate users. Removing this bar until a better solution is found.


cintek (changes requested):
These loading bars (when you open a source) should also be removed:
loading

cintek (approved):

PR #798, opened 2025-01-07
Display icons next to workflow states

cintek:
and use different icon for state "new" as the star looks more like "favorite" or "bookmark".

The icons next to the workflow states help users to understand the icons in the table (if they don't find the tooltips).

### DIFF
--- a/client/src/lib/Advisories/Advisory.svelte
+++ b/client/src/lib/Advisories/Advisory.svelte
@@ -391,7 +391,7 @@
       </Label>
     </div>
     <div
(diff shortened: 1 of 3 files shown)


koplas (approved):
LGTM

JanHoefelmeyer (dismissed):
The icon for new is hollow within the advisory view, but filled on the search table. (Regardless of light or dark mode)

This should be made consistent.

JanHoefelmeyer (approved):
LGTM & works

Issue #701, opened 2024-11-27
Clean Up permissions


JanHoefelmeyer:
On the backend reviewers can supposedly edit SSVCs, even though they shouldn't.

On the frontend, they can attempt but trying to save it will fail.

Currently, auditors can forward documents, which they shouldn't? (Not tested with frontend)

This should be looked at.

Other permissions might be similarly not as intended.

JanHoefelmeyer:
(Not allowing only-Auditors to forward documents since they are seen as passive users and are not supposed to change the system)

PR #803, opened 2025-01-08
Permissions cleanup

JanHoefelmeyer:
fixes https://github.com/ISDuBA/ISDuBA/issues/701

Permissions being sorted alphabetically (instead of not at all) should make them easier to maintain.

Removed Auditors ability to forward documents and Reviewers ability to set SSVC.


cintek (approved):
LGTM & works

PR #806, opened 2025-01-08
Dashboard: Show changes sources next to queries

cintek:
Before, the "Changed sources" on the dashboard were *always* below the other queries. Now they are next to them if there is enough space.


koplas (approved):
LGTM

PR #807, opened 2025-01-08
Extend OpenAPI documentation

koplas:
This pull request is only for better visibility. This will be later squashed and merged into `openapi`.


koplas:
Merged into openapi

PR #808, opened 2025-01-08
Fix layout problems

cintek:
When you open a source on a screen with low height and scroll to the bottom where is not space after the last text so it seems that you can scroll further down but you cannot:
layout

This PR re-adds padding at the bottom of the main content in the window. The removal was a workaround so the positioning of the diff box could work but now I found a different solution.


JanHoefelmeyer (approved):
works

PR #810, opened 2025-01-09
Modify default message, restrict to those it helps.

JanHoefelmeyer:
Admins cannot create (nor should they view) sources, so there's no reason to contact them if you want new ones. 
Similarly, there's not much of a reason to tell Source-Managers to contact themselves if something is missing.
(Importers also can upload whatever they may be missing.)

Also:
 * Should auditors be able to see (and thus audit) the sources tab? 
  -> After some consideration, yes they should

### DIFF
--- a/client/src/lib/SideNav.svelte
+++ b/client/src/lib/SideNav.svelte
@@ -86,7 +86,7 @@
                 <i class="bx bx-spreadsheet"></i>
               </svelte:fragment>
             </SidebarItem>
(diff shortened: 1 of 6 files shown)


JanHoefelmeyer:
Temporarily removing Editors ability to see Aggregators tab since it was not well documented and broken in multiple places

s-l-teichmann (approved):
LGTM

PR #816, opened 2025-01-10
Upload: Fix layout when list is very long

### LINKED ISSUE #815: import view: Border doesn't grow with list
In the upload view, the frame does not grow when multiple files are selected:

grafik

cintek:
Fixes #815.


koplas (approved):
LGTM

PR #824, opened 2025-01-15
Add coverage report with monocart-reporter

cintek:
This PR is using the [monocart-reporter](https://github.com/cenfun/monocart-reporter) to create coverage reports.

As you can see there are also source maps included in the report:
coverage

I created an [issue](https://github.com/cenfun/monocart-reporter/issues/165) to find out if they are counted to the overall coverage value (I guess they do) and how to prevent that.

At least the code from `node_modules` is filtered out.

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -39,12 +39,12 @@ jobs:
       - uses: actions/upload-artifact@v4
         if: always()
         with:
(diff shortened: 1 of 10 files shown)


koplas (dismissed):
LGTM

koplas (approved):
LGTM

PR #827, opened 2025-01-16
Coverage: Set logging level to "info"

cintek:
This will remove the source maps from the coverage report. See https://github.com/cenfun/monocart-reporter/issues/165#issuecomment-2592805714.


koplas (approved):
LGTM

PR #845, opened 2025-01-27
Tailwind 4 Migration

koplas:


bernhardreiter:
@koplas told me that this is a test and much is broken by the move to the new version. (So we do not want this for stable releases in the next weeks.)

JanHoefelmeyer (dismissed):
LGTM & works

JanHoefelmeyer (approved):

Issue #847, opened 2025-01-30
Misspelled workflow status


mfd2007:
Problem: Querying advisories with workflow status "archived" doesn't work.

Query `$state archived workflow = ` gives the following error:

> An error occured. 400: The request sent could not be understood. "archived" is not a valid workflow

Query `$state archive workflow = ` gives the following error:

> An error occured. 500: An error occured on the server. Please contact an administrator. cannot calculate count ERROR: invalid input value for enum workflow: "archive" (SQLSTATE 22P02)

Problem is a different spelled `archived` status here: https://github.com/ISDuBA/ISDuBA/blob/038e27e03659b2b17f8a1721d76b2ae032f3fba5/pkg/database/query/parser.go#L735

PR #848, opened 2025-01-30
fix: fix misspelled workflow status

mfd2007:
Change should harmonize the spelling of the workflow status "archived". This solves the #847

### DIFF
--- a/pkg/database/query/parser.go
+++ b/pkg/database/query/parser.go
@@ -732,7 +732,7 @@ func (*Parser) pushCmp(st *stack, et exprType) {
 
 var validWorkflows = []string{
 	"new", "read", "assessing",
-	"review", "archive", "delete",
+	"review", "archived", "delete",
 }
 
 func parseWorkflow(s string) string {


s-l-teichmann (approved):
LGTM

This is fine for now. A better way would be to use the centralized model
`Workflow` in https://github.com/ISDuBA/ISDuBA/blob/038e27e03659b2b17f8a1721d76b2ae032f3fba5/pkg/models/workflow.go#L18 and add `IsValid` method there. (which exists https://github.com/ISDuBA/ISDuBA/blob/038e27e03659b2b17f8a1721d76b2ae032f3fba5/pkg/models/workflow.go#L112 )This is currently not so easy .. mainly because we would introduce a circular import scenario (models <-> parser). This could be resolved by making e.g. `Expr` in interface in `models` and being `Expr` in `parser` an implementation for this. To do this some factory stuff must me introduced, plus some minor refactoring around the parser ... but all of this would be something for another PR.

PR #857, opened 2025-02-03
Group all documents

### LINKED ISSUE #800: `Search`: Many results of the same document take too much place
If you search for a term that appears multiple times in the same document you get a search result where each of these hits is displayed as a separate result. This way the results take much place. Here is an example where a document contains the term "download" three times and thus, the document is display three times in full height:
many-hits-one-document

To improve the usability hits in the same document should be summarized so it's easier to see that hits belong to the same document and the results don't take so much place.

koplas:
This prevents the same document from being displayed multiple times in the search.

TODO: Change the UI to accommodate the search results' grouping.

Closes #800

### DIFF
--- a/pkg/database/query/sqlbuilder.go
+++ b/pkg/database/query/sqlbuilder.go
@@ -17,12 +17,13 @@ import (
 
 // SQLBuilder helps to construct a SQL query.
 type SQLBuilder struct {
-	WhereClause  string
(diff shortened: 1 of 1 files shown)


koplas:
This has a high impact on performance if the search term occurs multiple times in the documents. Another approach is needed.

PR #864, opened 2025-02-07
Reset comment text area before sending request

### LINKED ISSUE #835: Enforce unique comments
On some machines there may be a delay before the send button greys out and the comment field empties when a comment is send. Clicking on send again before that happens will cause the comment to be send multiple times

koplas:
Closes #835


JanHoefelmeyer (approved):
lgtm & works

PR #877, opened 2025-02-10
Don't logout if there is still activity in other tabs

### LINKED ISSUE #867: Defect: Sudden logouts
I experienced sudden logouts while working with ISDuBA in multiple tabs. It is unclear to me, why it happened.

Steps to reproduce:
1. Use multiple tabs with different content from ISDuBA.
2. Work in one tab (for a longer time), don't touch the other.
3. During storing (SSVC, comment), the server replies with a 401 - not saving the changes made.

koplas:
Closes #867


bernhardreiter:
I wonder should we document our usage of the local browser storage?

JanHoefelmeyer (approved):

bernhardreiter:
> This is a technical detail and is not relevant to the user.

I would allow to estimate the attack surface and to explain why some browsers will log in and others do not in some situations. So it could be interesting to the operators.

PR #879, opened 2025-02-10
Add API to overview diagram

### LINKED ISSUE #837: Add API to overview diagram and link it from main Readme
Imports and other things can also happen directly on the server backend using the API. This should be added to the overview diagram.

The diagram should also be linked from the documentation, so that it is directly displayed. Main README?

cintek:
Closes #837.


s-l-teichmann (changes requested):
The API is used by the Web application so its between that and the backend.

s-l-teichmann:
> > The API is used by the Web application so its between that and the backend.
> 
> @s-l-teichmann I put it into a different place because you can also use it on the command-line via curl etc.

The Web App is using the API nonetheless. There is no other way to interact with the backend. Placing the API somewhere else suggests that there is something special in this communication ... which it isn't.

cintek:
Another try: This time I added "API" to the arrow between web application and backend.

s-l-teichmann:
> Another try: This time I added "API" to the arrow between web application and backend.

This is much better.

s-l-teichmann (approved):
LGTM

PR #883, opened 2025-02-11
fix: Restore remembered sorting

### LINKED ISSUE #805: Remember sorting of table
When the user changes the sorting of the table opens an advisory and goes back to the table the sorting is reset back to the default. This is annoying if the user wants to open several advisories.

cintek:
Fixes #805.


koplas (approved):
LGTM

PR #888, opened 2025-02-12
Show if source is active

### LINKED ISSUE #809: Aggregator view: Show if a source is active
In the aggregator view it should be visible if a source is active or not. Currently, if you have subscribed feeds of a source you might think that it is enabled though it's not.

cintek:
Closes #809.

With this PR the indicator which is used in the source list at `Sources` now is also used for sources inside the aggregator view:

active-source


koplas (approved):
LGTM & works

Issue #891, opened 2025-02-12
Fix Vulnerabilities/Scores/Score/CVSS V3 : ctx is not a function


JanHoefelmeyer:
Getting a 

Uncaught (in promise) TypeError: ctx[1][ctx[5]].startsWith is not a function

 error when trying to open

PR #892, opened 2025-02-12
Make parameter to string so it properly can be analyzed

JanHoefelmeyer:
Solves https://github.com/ISDuBA/ISDuBA/issues/891


JanHoefelmeyer:
superceeded by https://github.com/ISDuBA/ISDuBA/pull/893

PR #900, opened 2025-02-17
Add lock to aggregator toggle

### LINKED ISSUE #863: Defect: Aggregators do not extend sometimes
Sometimes the aggregator does not extend and there is no loading circle. A reload of the page helps.

Unfortunately, I can't reproduce this reliable. Therefore, the issue for now just documents the behavior and waits for others to provide more information and context on that bug.

_ISDuBA 0.7.3-1-gf61463de_

koplas:
If toggle is called multiple times and the previous request is slower than next, it will result in a race condition, where the state is not correctly set. The locking can be improved by introducing a separate lock per aggregator.

Closes #863

### DIFF
--- a/client/src/lib/Sources/Aggregators/AggregatorViewer.svelte
+++ b/client/src/lib/Sources/Aggregators/AggregatorViewer.svelte
@@ -348,27 +348,29 @@
   };
 
   const toggleAggregatorView = async (aggregator: Aggregator) => {
-    if (aggregator.id === undefined) {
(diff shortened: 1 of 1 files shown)


cintek:
> I couldn't reproduce the issue so I cannot say if this prevents the bug but using a lock looks like a good solution to me.

When I used throttling ("Good 3G" in Firefox) I could reproduce this and it's solved with this PR.

PR #903, opened 2025-02-18
fix: Increase visibility of indicator in dark mode

### LINKED ISSUE #901: Diff: Indicators almost invisible in dark mode
I think the screenshot makes clear what the problem is:
Image

cintek:
Fixes #901.


koplas (approved):
LGTM

PR #907, opened 2025-02-21
Add ilikepname expression

### LINKED ISSUE #774: Add query expression that allows to filter for product names
Currently, it is possible to filter for documents with specific product IDs using `ilikepid`. As product IDs don't need to contain the product's name, it can be helpful to filter for product names.

koplas:
Closes #774


JanHoefelmeyer (approved):
LGTM & works

PR #914, opened 2025-02-26
Correctly resolve url if base has no trailing slash

### LINKED ISSUE #912: ISDuBA can't retrieve Microsoft
We can't retrieve files from Microsoft (www.microsoft.com) right now. The `csaf_checker` however is able to retrieve the documents...

koplas:
If the URL was previously https://example.com/abc, the abc part was ignored.

Closes #912.

### DIFF
--- a/pkg/sources/indices.go
+++ b/pkg/sources/indices.go
@@ -39,7 +39,7 @@ func (fi *feedIndex) rolieLocations(r io.Reader) ([]location, error) {
 			return fmt.Errorf("invalid href: %v", href)
 		}
 		if !u.IsAbs() {
-			u = fi.base.JoinPath(u.Path)
+			u = joinURL(fi.base, u)
 		}
 		*store = u
 		return nil
@@ -141,7 +141,7 @@ func (fi *feedIndex) directoryLocations(r io.Reader) ([]location, error) {
 			return nil, fmt.Errorf("column 2 in line %d is not a valid RFC3339 time: %w", lineNo, err)
 		}
 		if !doc.IsAbs() {
-			doc = fi.base.JoinPath(doc.Path)
+			doc = joinURL(fi.base, doc)
 		}
 		// Apply age filter
 		if fi.age != nil && updated.Before(cut) {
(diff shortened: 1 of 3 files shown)


bernhardreiter:
Can you add a unit test as well?

pkg/sources/indices.go
```diff
@@ -141,7 +141,12 @@ func (fi *feedIndex) directoryLocations(r io.Reader) ([]location, error) {
 			return nil, fmt.Errorf("column 2 in line %d is not a valid RFC3339 time: %w", lineNo, err)
 		}
 		if !doc.IsAbs() {
-			doc = fi.base.ResolveReference(doc)
+			// Use copy to avoid side effects
+			base := *fi.base
+			if !strings.HasSuffix(base.Path, "/") {
```
s-l-teichmann, line 146: Should be better done in source.go before line 224 to move that out of the loop. That also would help the rolie path, which may be affected by this, too.

s-l-teichmann (changes requested):
See remarks

JanHoefelmeyer (approved):
Download works for me using this fix, testing using both Microsoft (which didn't work before) and BSI (which already worked before)

pkg/sources/indices.go
```diff
@@ -39,7 +39,7 @@ func (fi *feedIndex) rolieLocations(r io.Reader) ([]location, error) {
 			return fmt.Errorf("invalid href: %v", href)
 		}
 		if !u.IsAbs() {
-			u = fi.base.ResolveReference(u)
+			u = fi.base.JoinPath(u.Path)
```
s-l-teichmann, line 42: This may omit a potential query and/or fragment part.
`[scheme:][//[userinfo@]host][/]path[?query][#fragment]`

s-l-teichmann (changes requested):

s-l-teichmann (changes requested):
LGTM (beside remarks)

pkg/sources/indices.go
koplas: Done

pkg/sources/util.go
```diff
@@ -45,3 +46,13 @@ func AsRegexps(s []string) ([]*regexp.Regexp, error) {
 	}
 	return slice, nil
 }
+
+// JoinURL joins the two URLs while preserving the query and fragment part of the latter.
```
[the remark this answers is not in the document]
koplas: Done

s-l-teichmann (approved):

PR #921, opened 2025-02-27
Improve error log streaming

s-l-teichmann:
The previous version flushes the stream for every entry.
This is not really needed for a download.
Implementing a Gin  renderer reduces the number of calls for each log line.
The context of the http request is now passed to the database operations so they get cancelled if the request is terminated early

### DIFF
--- a/pkg/web/sources.go
+++ b/pkg/web/sources.go
@@ -809,7 +809,9 @@ func (lr *logRenderer) Render(w http.ResponseWriter) error {
 	already := false
 	for entry := range lr.entries {
 		if already {
-			fmt.Fprint(w, ",")
+			if _, err := fmt.Fprint(w, ","); err != nil {
+				return err
+			}
 		} else {
 			already = true
 		}
@@ -906,6 +908,7 @@ func (c *Controller) feedLogs(ctx *gin.Context, feedID *int64) {
 		return
 	}
 
+	ctx.Header("Content-Type", "application/json")
 	ctx.Render(http.StatusOK, &lr)
 }
 
--- a/pkg/web/sources.go
+++ b/pkg/web/sources.go
@@ -13,7 +13,6 @@ import (
 	"encoding/pem"
 	"errors"
 	"fmt"
-	"io"
 	"iter"
 	"log/slog"
 	"net/http"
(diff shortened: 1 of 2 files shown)


koplas (changes requested):
I tried downloading feed logs with the UI. The resulting file looks like this:
```JSON
[null,0]
```
The Feed Logviewer also displays no feed log entries. This does not happen on the main branch.

pkg/web/sources.go
```diff
@@ -783,6 +782,50 @@ func (c *Controller) allFeedsLog(ctx *gin.Context) {
...
+	} else {
+		_, err = fmt.Fprint(w, `{"entries":[`)
+	}
+	if err != nil {
+		return nil
+	}
+	already := false
+	for entry := range lr.entries {
+		if already {
+			fmt.Fprint(w, ",")
```
koplas, line 812: The error here is unhandled.

pkg/web/sources.go
```diff
@@ -841,48 +884,29 @@ func (c *Controller) feedLogs(ctx *gin.Context, feedID *int64) {
...
 		slog.Error("database error", "error", err)
 		models.SendError(ctx, http.StatusInternalServerError, err)
 		return
 	}
 
-	next, stop := iter.Pull(stream)
-	defer stop()
-	var num int64
-	ctx.Writer.Header().Add("Content-Type", "application/json")
-	ctx.Stream(func(w io.Writer) bool {
```
koplas, line 859: `Stream` seems like the correct way to implement this.

pkg/web/sources.go
s-l-teichmann: This is indented.

pkg/web/sources.go
s-l-teichmann: Stream is more for thinks like Server Sent Events and is flushing after each write.

s-l-teichmann:
> I tried downloading feed logs with the UI. The resulting file looks like this:
> 
> ```json
> [null,0]
> ```
> 
> The Feed Logviewer also displays no feed log entries. This does not happen on the main branch.

Should be fixed now.

koplas (approved):
LGTM

Issue #988, opened 2025-05-15
Definition mismatch of duration enums


wagner-intevation:
The API call `/api/client-config` on a demo instance returns
```json
{
  "keycloak_url": "...",
  "keycloak_realm": "isduba",
  "keycloak_client_id": "auth",
  "update_interval": 300000000000,
  "idle_timeout": 1800000000000
}
```

However the definition of the call says:

https://github.com/ISDuBA/ISDuBA/blob/0fc75557c227d18da0d4148c066ca52e9a6dfe51/pkg/web/docs/docs.go#L3439-L3458

and `time.Duration` is defined as

https://github.com/ISDuBA/ISDuBA/blob/0fc75557c227d18da0d4148c066ca52e9a6dfe51/pkg/web/docs/docs.go#L3824-L3835

which allows neither 300000000000 nor 1800000000000

-> the response from isduba is invalid according to it's own API definition

koplas:
Fixed with #989.

PR #989, opened 2025-05-15
Use integer type for timestamps

koplas:
This avoids generating client code that does not accept valid values.


wagner-intevation (approved):
I can confirm it fixes https://github.com/ISDuBA/ISDuBA/issues/988

bernhardreiter:
This one did not have a code review, but quite a few code changes without detailed explanation.
@s-l-teichmann can you do a second review before we release.

PR #1019, opened 2025-06-11
Add health indicator

koplas:


cintek (approved):
LGTM on the client side & works

pkg/web/sources.go
```diff
@@ -132,11 +139,53 @@ func showStats(ctx *gin.Context) (bool, bool) {
...
 
+func showHealth(ctx *gin.Context) (bool, bool) {
+	st := ctx.Query("health")
+	if st == "" {
+		return false, true
+	}
+	return parse(ctx, strconv.ParseBool, st)
+}
+
+func (c *Controller) isHealthy(ctx context.Context, isSource bool, id int64) (*bool, error) {
```
s-l-teichmann, line 150: Return a simple bool instead of *bool here because the NOT EXISTS query always returns a bool value. The pointers upwards the stack could therefore also replaced by a simple bool.

s-l-teichmann (changes requested):

JanHoefelmeyer (dismissed):
lgtm & works

s-l-teichmann (approved):
LGTM

PR #1032, opened 2025-06-26
Only configure timeout on config load

koplas:
This avoid accessing data that not is not available.


JanHoefelmeyer (approved):
LGTM & works

PR #1039, opened 2025-06-27
Fix a query order race condition

koplas:
After a user cloned a query, there is a time window where the client requests the current query data, and the current client side data is stale. When the user changes the stale query order it will conflict with the current data.
This can happen on slow or bad connections.


JanHoefelmeyer (approved):
LGTM & works

PR #1040, opened 2025-06-27
Try the previous page if there are no documents on the current

koplas:
This avoids breaking the navigation when documents are deleted, while the user is on the last page.
on the last page.


JanHoefelmeyer (approved):
LGTM & works

PR #1043, opened 2025-07-01
fix: Prevent errors

cintek:
Before, there was a case where productNames was false and the application became unresponsive. While at it also prevented the error in ProductGroups.

This doesn't solve the cause of the problem which needs more investigation.


JanHoefelmeyer (approved):
LGTM & works

Issue #849, opened 2025-01-30
Search query not completely visible


mfd2007:
Problem: if a longer search string is entered, the end is hidden by the `apply` button.

![Image](https://github.com/user-attachments/assets/5736b278-289e-4f21-bab3-b534fa55c5a3)

Expectation: Move the `apply` button next to the input field.

JanHoefelmeyer:
Solved via #1059

PR #1059, opened 2025-07-10
Fix: Search apply button no longer in the search bar prevents it from obstructing search queries.

JanHoefelmeyer:
Solves https://github.com/ISDuBA/ISDuBA/issues/849


koplas (approved):
LGTM

Issue #967, opened 2025-04-17
Statistic diagrams show time not time instead of date


mfd2007:
If a time period of less then 16 day is selected, the diagram shows only the and not the date.

Time period 16 days

![Image](https://github.com/user-attachments/assets/98353cfa-551b-4e3c-8611-2e082b191430)

Time period 17 day

![Image](https://github.com/user-attachments/assets/908be5fa-dc28-425a-94c0-08aa32756439)

Would be more convenient to see the date, when a days are selected.

PR #1060, opened 2025-07-10
Fix: Change granularity of time axis depending on size of steps on X axis.

JanHoefelmeyer:
Solves https://github.com/ISDuBA/ISDuBA/issues/967


koplas (approved):
LGTM & works

Issue #1000, opened 2025-05-20
`Draft` should be displayed before `final`


tschmidtb51:
Currently, it can happen that multiple documents of the same advisory exist. The draft version is displayed after final version, which does not make sense.

PR #1061, opened 2025-07-10
Now also sort by tracking status, create Type for it to guarantee values

JanHoefelmeyer:
Solves https://github.com/ISDuBA/ISDuBA/issues/1000


koplas (approved):
LGTM

PR #1083, opened 2025-08-15
Add previous and next advisory buttons

koplas:
The search query and offset are stored when opening an advisory. This is used to retrieve the neighboring advisories.
Currently the browser history is not considered.


koplas:
How to reproduce the issue with the history:
- Open a advisory
- Press `Next advisory`
- Press browser back button
- Press `Next advisory`
- Press `Previous advisory`

If this works you should land on the first advisory. This does not currently happen.

koplas:
It could be enough to listen for the [popstate](https://developer.mozilla.org/en-US/docs/Web/API/Window/popstate_event) event and clear the SessionStorage entry. Pressing one time on the back button would then disable the buttons.

PR #1090, opened 2025-08-18
Display error for individual files

koplas:
Don't abort when a single error occurs and display the success status for each individual file.


JanHoefelmeyer (approved):
LGTM & works

Issue #976, opened 2025-05-06
Improve manual import


tschmidtb51:
When doing an manual import, it is unclear which documents are already processed and when the process is finished. It is also hard to figure out which document raised the error if one occurred.

tschmidtb51:
@bernhardreiter Please add that to the list of estimations

tschmidtb51:
Maybe, we could add a green check mark / red x / yellow load to indicate the status

PR #1119, opened 2025-09-09
Improve manual imports feedback

cintek:
Solves #976.

### DIFF
--- a/client/src/lib/Sources/DocumentUpload.svelte
+++ b/client/src/lib/Sources/DocumentUpload.svelte
@@ -14,13 +14,17 @@
   import Upload from "$lib/Upload.svelte";
   import { request } from "$lib/request";
 
+  let cancelled = false;
+
   const uploadDocuments = async (
     files: FileList,
     updateCallback: ((uploadInfo: UploadInfo[]) => void) | undefined
   ): Promise<UploadInfo[]> => {
+    cancelled = false;
     if (!files) return [];
     let uploadInfo = [];
     for (const file of files) {
+      if (cancelled) break;
       let info: UploadInfo = { success: true };
       const formData = new FormData();
       formData.append("file", file);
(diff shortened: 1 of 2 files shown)


cintek:
- Upload button is now at the top (in case the list is very long the user doesn't have to scroll a long way)
- Error messages are visible right away (no tooltip anymore)

list
list-with-errors

cintek:
The progress is shown now next to the button:
progress

But if you upload just one or a few documents the text might disappear so fast that the user is confused. Maybe we should just show a spinner when there are <20 (or a different limit) documents to upload.

JanHoefelmeyer (dismissed):
LGTM & works

bernhardreiter (dismissed):
* Filenames in the table are almost invisible in dark mode.

grafik

The `undefined` in the messages looks strange, e.g. in `Could not upload file 409: An error occured. Please contact an administrator. already in database undefined`.

Why should I contact an "administrator" when trying to upload a file that is already there. Can we adapt that message?

Pressing "upload" with no file selected gives an error on the console and shows `Uploading ...
0/1` forever ....

cintek:
> Can we do an automatic test for at least a manual upload easily?

There was already one. I extended it so it also checks if the button is disabled when no files are selected.

cintek:
> @cintek can be also remove the list of files, if a new set of files was selected?
> 
> Currently the list of previously imported files still is there (without status) even when I select new different files.


Now the list will be updated when new documents are selected

JanHoefelmeyer (approved):
LGTM & works, but:
Maybe add a small spinner when files are loaded after selecting files.

With many files the loading time is long enough to require that.

cintek:
> Maybe add a small spinner when files are loaded after selecting files.
> 
> With many files the loading time is long enough to require that.

I'm not sure how to accomplish this. It is possible to listen for the `click` event of the file input but when the spinner is already visible while the file dialog is open it is confusing for the user. They might think there is something else going on in the background.

The `input` event is never fired and the `change` event when the files are already loaded into the browser. So there events cannot be used to achieve this.

But because we were writing about many files I implemented a button to cancel the upload.

Issue #1142, opened 2025-09-29
Queries with criteria not saved when switching to incompatible type


cintek:
Steps to reproduce: 
- Create a query (type: `Advisories`, query criteria: `$state review workflow =`)
- Open the query, set type to `Documents` and empty query criteria (you also have to select at least one column of course because the selection is reset when the type is changed)
- Save the query

Expected: The query was saved with the updated attributes.

Result: The query has still the old attributes.

PR #1158, opened 2025-10-14
fix: Update query without criteria

cintek:
Solves #1142.


JanHoefelmeyer (changes requested):
~~This seems to also catch if a wrong query is used. So any typo or misuse of the query data is saved as simply "True". The error message will appear for a splitsecond before the user is send back to the query overview, but unless you're actively looking for it it's easy to miss. This will give the wrong impression that a filter was used.~~

^ Is *not* saved, also, is an issue in main as well, unrelated to this one. This PR works

Issue #1138, opened 2025-09-25
Can't set dashboard query after reordering queries


JanHoefelmeyer:
"Could not change option. 400: The request sent could not be understood. not a unique value: %sFEHLER: doppelter Schlüsselwert verletzt Unique-Constraint »stored_queries_definer_num_key« (SQLSTATE 23505)" when trying to set a dashboard query,  after reordering queries.

PR #1160, opened 2025-10-14
fix: Set dashboard after queries were reordered

cintek:
Solves #1138.


bernhardreiter:
> Solves https://github.com/ISDuBA/ISDuBA/issues/1138.

in our current setting this seems to only create a _mention_. So we need to set the _this merge request may close_ manually, which I now did.

JanHoefelmeyer (approved):
LGTM & works

Issue #1077, opened 2025-07-30
test and coverage for 0.11.0


bernhardreiter:

JanHoefelmeyer:
As for the server-side
```go test -cover ./...
	github.com/ISDuBA/ISDuBA/cmd/bulkimport					coverage: 0.0% of statements
	github.com/ISDuBA/ISDuBA/cmd/isdubad		       		coverage: 0.0% of statements
	github.com/ISDuBA/ISDuBA/pkg/aggregators				coverage: 0.0% of statements
	github.com/ISDuBA/ISDuBA/pkg/cache						coverage: 0.0% of statements
	github.com/ISDuBA/ISDuBA/pkg/config						coverage: 0.0% of statements
	github.com/ISDuBA/ISDuBA/pkg/database					coverage: 0.0% of statements
ok  	github.com/ISDuBA/ISDuBA/pkg/database/query	0.010s	coverage: 7.4% of statements
	github.com/ISDuBA/ISDuBA/pkg/forwarder					coverage: 0.0% of statements
	github.com/ISDuBA/ISDuBA/pkg/ginkeycloak				coverage: 0.0% of statements
ok  	github.com/ISDuBA/ISDuBA/pkg/models			0.014s	coverage: 25.5% of statements
	github.com/ISDuBA/ISDuBA/pkg/sources					coverage: 0.0% of statements
	github.com/ISDuBA/ISDuBA/pkg/tempstore					coverage: 0.0% of statements
?   	github.com/ISDuBA/ISDuBA/pkg/version					[no test files]
	github.com/ISDuBA/ISDuBA/pkg/web						coverage: 0.0% of statements
	github.com/ISDuBA/ISDuBA/pkg/web/docs					coverage: 0.0% of statements
```

PR #1164, opened 2025-10-15
Add more tests for client

cintek:
Part of #1077.

When coverage is mentioned in this PR only the coverage of the client is meant.

Coverage on main (2025-10-15):

- Bytes: 46.48%
- Statements: 17.62%
- Branches: 15:70%
- Functions: 16.64%
- Lines: 46.85%


cintek:
> I take it that is the client coverage?

Yes, I changed the title and the description to reflect that fact.

PR #1173, opened 2025-10-20
feat: Show message if URL is inconsistent

### LINKED ISSUE #1161: URL Manipulation results in inconsistent document view
While looking for a way to generate view-URLs from API-URLs I noticed an issue with how ISDuBA handles view urls.

The ISDuBA user interface generates URLs like such:

`http://localhost:5371/#/advisories/Red Hat Product Security/CVE-2024-32962/documents/42090`

This gives the following display:

Image

The 42090 is the internal document ID that can also be found in API calls, like
`http://localhost:5371/api/documents/42090`

If I manipulate the URL, and change the document ID to something else:
`http://localhost:5371/#/advisories/Red Hat Product Security/CVE-2024-32962/documents/12`

I get the following:

Image

The top still states I'm looking at **CVE-2024-32962** from **Red Hat**, but the document below it is for **CVE-2022-1048** of **BSI**. In other words, the GUI blindly takes the data from the URL, without checking if this data is consistent. Only when I use a non-existing Publisher or a non-existing CVE does the document not load. Though it fails to load because of an internal JavaScript error, not because it validated the data.

cintek:
Resolves #1161.

### DIFF
--- a/client/src/lib/Advisories/Advisory.svelte
+++ b/client/src/lib/Advisories/Advisory.svelte
@@ -28,6 +28,7 @@
   import SsvcBadge from "./SSVC/SSVCBadge.svelte";
   import { addSlashes } from "$lib/utils";
   import type { TrackingStatus, AdvisoryVersion } from "./advisory.ts";
+  import InconsistencyMessage from "$lib/Advisories/InconsistencyMessage.svelte";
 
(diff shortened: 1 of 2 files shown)


JanHoefelmeyer (approved):
LGTM & works

PR #1206, opened 2025-11-03
Prevent forbidden requests

### LINKED ISSUE #1159: Prevent forbidden requests to `sources/attention`
When the user is on the dashboard the browser sends requests to `sources/attention` even if the user is not allowed to receive any information from this endpoint:

Image

cintek:
Resolves #1159.


JanHoefelmeyer (approved):
LGTM & works

PR #1222, opened 2025-12-02
Per default, set up less users to reduce setup time

JanHoefelmeyer:
This should lower the setup time for testing etc.

### DIFF
--- a/docs/developer/users.txt
+++ b/docs/developer/users.txt
@@ -7,24 +7,24 @@
 # Software-Engineering: 2024 Intevation GmbH <https://intevation.de>
 
 # Username  First_Name  Last_Name      Email-Adress       Password Group      Role
-  beate     Beate       Bear           bea@example.com    beate    all        editor
-  revy      Revy        Reviewer       rev@example.com    revy     all        reviewer
-  imke      Imke        Importer       imk@example.com    imke     all        importer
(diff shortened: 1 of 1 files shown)


JanHoefelmeyer:
This is disabling all users except for the test-user.
The unaltered users.txt file is designed to set up users with a single role. The test-user has all roles, which is why they were set up manually after the users.txt users were created.

Unit tests and most manual tests  only need the test-user, but currently create all users.txt users for even the slightest changes. Unit-tests take around 10 minutes, with a significant part of it taken up by the user-creation process. For reference, the unit test of this Pull-request took only 6 minutes and 50 seconds, which is a significant time saver. This PR will make future tests and testing much faster. 

If the other users should be created, editing the users.txt is simple, and then the relevant scripts can simply be called again.

However, this was a minimalist solution to the problem. I do see the benefits of making the number and complexity of users created by the setup script configurable, so unit tests could simply go with the minimalist approach, while someone who wants to thoroughly test the application could simply use the full setup without having to manually edit the users.txt.

JanHoefelmeyer:
To further quicken the setup process, groups and roles could be created behind a similar logic, possibly saving even more time

JanHoefelmeyer:
superceeded by #1232

PR #1254, opened 2026-01-14
Update docs/scripts for dev installation, to nodejs 24 and more

bernhardreiter:
* update keycloak version to 26.5.0 from 26.0.0
 * update nodejs version to 24 from 20 (as support for 20 runs out in 2026-04)
 * remove superfluous `npm run prepare` as this is run by `npm install` already.
 * update how installpostgres.sh installs the GnuPG pubkey for the Postgresql apt repo.
   (As apt-key is deprecated.)

### DIFF
--- a/docs/scripts/installkeycloak.sh
+++ b/docs/scripts/installkeycloak.sh
@@ -12,7 +12,7 @@ set -e # to exit if a command in the script fails
 
 sudo apt install -y unzip # needed to unzip the keycloak archive
 
-version="26.0.0"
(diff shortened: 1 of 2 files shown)


bernhardreiter:
We should check if node20 is used else where.
I could build eveything, start isduba, log-in with test-user and add intevation as source and download the 5 documents.

cintek:
> @cintek can you check why the playwrighttest fails?

Hard to say why it failed but the second run was successful.

PR #1257, opened 2026-01-15
Fix: Application of classes for table body cells

cintek:
This PR fixes classes for table cells. Partly layout issues are remainder of flowbite-svelte migration and partly probably because of a replace-all-action. Then this PR applies classes where they were missing until now for more consistency.


JanHoefelmeyer (approved):
Works & LGTM

PR #1323, opened 2026-02-23
Make SSVC badge look like label

cintek:
Buttons are rounded and labels beside the SSVC badge are not so the badge had to be changed to be consistent with other labels.

Also make the badge smaller.

Before:

Screenshot from 2026-02-23 14-16-33

After:

Screenshot from 2026-02-23 14-17-35


bernhardreiter (approved):
Good enough

Issue #1317, opened 2026-02-16
Tests fail because Upload button is disabled


cintek:
Today a [test workflow](https://github.com/ISDuBA/ISDuBA/actions/runs/22053385290?pr=1313) failed because of the error below. Recently, @bernhardreiter had the the same error on his VM.

```
Error: [locator.click](http://locator.click/): Test timeout of 30000ms exceeded.
Call log:
  - waiting for getByRole('button', { name: 'Upload' })
    - locator resolved to <button disabled type="button" class="text-center font-medium inline-flex items-center justify-center text-white bg-primary-700 hover:bg-primary-800 dark:bg-primary-600 dark:hover:bg-primary-700 focus-within:ring-primary-300 dark:focus-within:ring-primary-800 px-5 py-2.5 text-sm focus-within:ring-4 focus-within:outline-hidden cursor-not-allowed opacity-50 rounded-lg">…</button>
  - attempting click action
    2 × waiting for element to be visible, enabled and stable
      - element is not enabled
    - retrying click action
    - waiting 20ms
    2 × waiting for element to be visible, enabled and stable
      - element is not enabled
    - retrying click action
      - waiting 100ms
    51 × waiting for element to be visible, enabled and stable
       - element is not enabled
     - retrying click action
       - waiting 500ms


  58 |     buffer: Buffer.from(arrayBuffer)
  59 |   });
> 60 |   await uploadButton.click();
     |                      ^

    at /home/runner/work/ISDuBA/ISDuBA/client/tests/setup.ts:60:22
```

PR #1327, opened 2026-02-23
Disable file input while documents are uploaded

cintek:
This **might** solve #1317 because the setup test will probably have to wait until the upload is finished until it can call `setInputFiles` again. And if it waits then the upload button should be enabled the second time.

Since we are also disable the upload button while the files are uploaded it is consistent to disable the file input too.


JanHoefelmeyer (approved):
LGTM && works

PR #1343, opened 2026-02-27
fix client vulnerabilities

JanHoefelmeyer:
Removes all current client vulnerabilities that npm warns us about


bernhardreiter:
Just a remark: as we do not know if the "weaknesses" of those dependencies will actually make ISDuBA vulnerable, we should call them "potential vulnerabilities" in dependencies. Or just "upgrading" client dependencies.

Background: As we have a security border between server and SPA client, most of the problems of the dependencies will not affect us.

Issue #1360, opened 2026-03-11
Search in Events tab is broken


JanHoefelmeyer:
Searching for any string via the non-advanced text search only wields a database error: Could not load query. 500: An error occured on the server. Please contact an administrator. cannot calculate count FEHLER: Syntaxfehler bei »)« (SQLSTATE 42601)

bernhardreiter:
Should be fixed with 0.14.0.  A brief test with 0.14.1-6-gf1655ea3 cannot reproduce the issue.

Issue #1349, opened 2026-03-03
Backend (postgres) Search timeout


JanHoefelmeyer:
When searching through many advisories, sometimes the search will timeout, which is not nice, especially as the number of advisories keeps increasing.

bernhardreiter:
https://github.com/ISDuBA/ISDuBA/releases/tag/v0.14.0 has a number of improvements for this problem.

Especially a overhauled SQL query builder. 
However the switch to the new code is not complete (#1403) and the new code misses some features (e.g. #1415)

tschmidtb51:
Please provide an estimation for this (including the sub-issues).

bernhardreiter:
> Please provide an estimation for this (including the sub-issues).

We already did a coarse one (see internal information).

PR #1367, opened 2026-03-12
Temporary event search fix

JanHoefelmeyer:
Solves https://github.com/ISDuBA/ISDuBA/issues/1360

When searching events for text now, this implementation returns every event related to an advisory where the text was found in the advisory or in comments. 

https://github.com/ISDuBA/ISDuBA/issues/1349 might solve this more gracefully, but for now this is usable and prevents an error.


s-l-teichmann (approved):
Okay as a temp solution. -> LGTM

PR #1382, opened 2026-03-13
fix: SSVC calculator layout

cintek:
Not all options of a complex decision were visible right away.

Before (you had to scroll to see the last option):
Screenshot from 2026-03-13 16-24-45

After:
Screenshot from 2026-03-13 16-24-56


JanHoefelmeyer (approved):
LGTM & works

PR #1394, opened 2026-03-23
Refactor querybuilder

s-l-teichmann:
This should fix some of the performance issue related to common full text searches.


s-l-teichmann:
I've switched it back to draft as the text EXITS checks have to be converted to JOIN CROSS LATERALs.

PR #1420, opened 2026-04-01
Fix CTE mode ilikepid/ilikepname

s-l-teichmann:
This should make the ilikepid and ilikepname working again.


JanHoefelmeyer (approved):
Works

PR #1429, opened 2026-04-07
Remove hover effect of badge

### LINKED ISSUE #1411: Aggregator: Text of badge barely readable on hover
The text inside the dismissable badge informing about changes of the aggregator is barely readable when hovering over it in dark mode.

Image

cintek:
Resolves #1411.

The hover effect has only the disadvantage that the text in barely readable in dark mode and it makes the user think that the whole badge is a button which is not the case.

Also, remove some props of `CBadge` we don't need.


JanHoefelmeyer (approved):
LGTM & works

PR #1432, opened 2026-04-13
fix: Values of CPE and purl

cintek:
Resolves #1431.

PR #1434, opened 2026-04-13
Add test for purl

cintek:
This will only run successful if #1432 is merged.


cintek:
Playwright test succeeds: https://github.com/ISDuBA/ISDuBA/actions/runs/24836416260

JanHoefelmeyer (approved):
LGTM & works

PR #1444, opened 2026-04-17
Display matches within one document

### LINKED ISSUE #291: Search: Better display of several matches within on document/advisory
When searching for something, for example a string, and there are several hits within one document  or one advisory, we display the document once per hit.

This clutters up the screen as the next document with a hit will potentially be off-screen initially. And the user would need to scroll down 20 hits.

An improvement would be to aggregate the hits into one document (or advisory) line which also gives the number of hits and potentially a few significant samples with context. 

If #106 is inplemented in the advisory details view, the search view can show less. (But the two issues can be implemented separately from each other.)

cintek:
Resolves #291.


JanHoefelmeyer (approved):
LGTM & works

PR #1447, opened 2026-04-22
fix: SSVC entries of other versions were not shown

### LINKED ISSUE #1446: SSVC values not shown in history for other document revisions
The SSVC of other document "versions" is not shown, reproducted with Version: 0.14.1-6-gf1655ea3

### Reproduction
* select an advisory that has several document versions (e.g. use sorting of VERSIONS to find one).
* enter an ssvc, switch to history, it can be seen
* select a different version of the document, enter an ssvc, switch to history, it can be seen but the other does isn't.
* switch to a completely different document version and history: now both new ssvcs are not seen.

See screenshot: 

Image

Expected: SSVCs are visible

cintek:
Resolves #1446.

### DIFF
--- a/client/src/lib/Advisories/Advisory.svelte
+++ b/client/src/lib/Advisories/Advisory.svelte
@@ -285,8 +285,12 @@
         }
       }
       if (e.event_type === "add_sscv" || e.event_type === "change_sscv") {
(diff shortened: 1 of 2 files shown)


cintek:
Playwright test: https://github.com/ISDuBA/ISDuBA/actions/runs/24796174902

JanHoefelmeyer (approved):
LGTM & works

PR #1466, opened 2026-04-29
fix: Tracking status not always shown for same version

cintek:
When the doc where the users comes from had the same version as a related doc the tracking status wasn't displayed although it should to be consistent:
Screenshot from 2026-04-29 08-54-30

With this PR:
Screenshot from 2026-04-29 08-57-12


JanHoefelmeyer (approved):
LGTM & works

PR #1469, opened 2026-04-29
Downgrade pgx PostgreSQL driver to 5.9.1

s-l-teichmann:
There is an issue with type casts in the latest version of the pgx driver.
We need to analyze this further. Till then go to the version before.


JanHoefelmeyer (approved):
LGTM & works

Issue #1470, opened 2026-04-29
documents with invalid JSON in the database


bernhardreiter:
One of our instances has documents that are not valid JSON as they contains bytes that are invalid in UTF-8.

Example opensuse-su-2025_14737-1.json in the database has hex `a0a0` inside.

Detection method (very simple for the first problematic document), run in pqsql:
```
DO $$
DECLARE
  orig RECORD;
  result jsonb;
BEGIN
    FOR orig IN
        SELECT filename, original FROM documents LIMIT 500
    LOOP
        RAISE NOTICE '%', orig.filename;
        result = convert_from(orig.original, 'UTF8');
    END LOOP;
    RAISE NOTICE '%', 'Done';
END; $$;
```

Error message is like
```
NOTICE:  suse-su-2023_4343-1.json
ERROR:  invalid byte sequence for encoding "UTF8": 0xa0
```

Expected behaviour: CSAF documents with invalid JSON should not be imported into the database.

Reproducing:
* curl -v -O https://ftp.suse.com/pub/projects/security/csaf/opensuse-su-2025_14737-1.json does _not_ have the problematic characters at time of writing , but four spaces in their place.

#### how to display a file in psql
```sql
SET bytea_output = 'escape';
 select original from  documents where filename = 'opensuse-su-2025_14737-1.json';
```

#### how to copy a bytea to file
This does not work directly, the following psql command will write out the value, but the header and footer (2 bytes) will have to be cut:

```
\copy (select original from  documents where filename = 'opensuse-su-2025_14737-1.json') TO 'o' WITH BINARY;
```

bernhardreiter:
Here is an example that suse.com distributed before. (They seem to have silently fixed the issue, but did not add a revision to the document.)

[opensuse-su-2025_14737-1.json](https://github.com/user-attachments/files/27204610/opensuse-su-2025_14737-1.json)

[opensuse-su-2025_14737-1.json.asc.txt](https://github.com/user-attachments/files/27204636/opensuse-su-2025_14737-1.json.asc.txt)

With https://ftp.suse.com/pub/projects/security/keys/security-automation.asc someone can verify the signature which shows that the attached file was distributed by suse.com.

```bash
gpg --fetch-keys https://ftp.suse.com/pub/projects/security/keys/security-automation.asc
gpg --verify opensuse-su-2025_14737-1.json.asc.txt opensuse-su-2025_14737-1.json
[...]

PR #1476, opened 2026-04-29
Collect bad strings on import and return an error if there are any.

s-l-teichmann:
Helps with #1470  by rejecting their import.


bernhardreiter:
Parts of the reasoning in the first answer of https://github.com/ISDuBA/ISDuBA/issues/1470#issuecomment-4350506888 should go into the code as comment or so, to document why valid incoming CSAF documents MUST not contain a unicode replacement character. This makes this fix a good one.

In addition isn't this a good opportunity for a solitary unit test?

s-l-teichmann:
@JanHoefelmeyer Can you check if this works by trying to upload a bad document which contains strings that are not properly UTF-8 encoded strings? BTW: The bulkimporter uses the same code path.

s-l-teichmann:
@bernhardreiter  Added a comment about the decoding error detection.

In general I agree that there should be more testing on the import code but
testing this code path would be oddly specific. To make this code better suited
for testing we would need to refactor it a bit like separating the PostgreSQL
from the JSON import part. I really would like to do this in a later PR.

IMHO we shouldn't delay this fix to go into the upcoming release to prevent
the databases to get more tainted as they are already are.

bernhardreiter (approved):
LGTM

JanHoefelmeyer (approved):
Works

PR #1479, opened 2026-04-30
fix: Bulk-change workflow states

### LINKED ISSUE #1473: Search Page: Bulk state change greyed out  no matter what
Even with correct permissions, even when changing just a single advisory, it's greyed out.

cintek:
Resolves #1473.


cintek:
Playwright test: https://github.com/ISDuBA/ISDuBA/actions/runs/25151423092

JanHoefelmeyer (approved):
LGTM & works

PR #1486, opened 2026-05-05
Improve sources test

cintek:
When I took a look at the report of one failed test I saw that the name of a source was not saved correctly. This should only be a problem in the automatic test but I fixed it for better quality.


cintek:
Playwright test: https://github.com/ISDuBA/ISDuBA/actions/runs/25369243548

JanHoefelmeyer (approved):
LGTM

PR #1488, opened 2026-05-06
codequality: add comments to the sql schema

bernhardreiter:
to explain a bit more how texts are stored into documents.document.


bernhardreiter:
@s-l-teichmann not perfect, but would have been a help, and could be removed if we have more documentation elsewhere.

s-l-teichmann (approved):
Not a huge improvement but if it helps ...

PR #1491, opened 2026-05-07
Return event count if requested @ 0 results

### LINKED ISSUE #1490: Query editor: Wrong result when backend returns `null`
When the user creates an event query, tests it and the backend returns `{"events":null}` the client shows the message "The query found results." Obviously this is not correct.

JanHoefelmeyer:
Resolves https://github.com/ISDuBA/ISDuBA/issues/1490


cintek (approved):
Works

PR #1494, opened 2026-05-07
Improve workflow state buttons

cintek:
Improve maintainability:

- less props have to be passed
- variables and methods are less complicated (e.g. `buttonClass` is now a simple string)

Improve UI when advisory view is loaded:

The workflow state buttons are now visible right away but they are disabled until the advisory state is known. This way at least this part of the UI is less "noisy".


cintek:
PW-Test: https://github.com/ISDuBA/ISDuBA/actions/runs/25513115399

JanHoefelmeyer (approved):
LGTM & works

PR #1495, opened 2026-05-08
fix: Match list style in Search

cintek:
The rows of the matches below the documents did not reach to the end so you had a gap at the right side:

match-list


JanHoefelmeyer (approved):
LGTM & works

PR #1500, opened 2026-05-08
Improve PrevNext

### LINKED ISSUE #1489: PrevNext disappears in some cases
- Set `Advisories per page` to more than 10
- Open the 11th document or one of the following on the same page
- Switch to the next document

Result: PrevNext is gone

Expected: PrevNext should still be visible

The problem is that in PrevNext the same `offset` as in Search is used but the `limit` is always set to 10 because we don't want to load to many results. We have to adjust the `offset` when the user selects a document that is further away from the start of the list.

### LINKED ISSUE #1504: Prev/Next: Navigating too fast can cause inconsistent states
You may end up at nonexistent advisories, or glitched states.

cintek:
Don't force user to wait until results are loaded. Abort old requests.
Resolves #1504.
Resolves #1489.

### DIFF
--- a/client/src/lib/Advisories/Advisory.svelte
+++ b/client/src/lib/Advisories/Advisory.svelte
@@ -69,6 +69,7 @@
   let couldNotLoadDocument = $state(false);
   let relatedDocuments: any = $state(undefined);
   let isLoadingSearchMatches = $state(false);
+  let abortControllers: AbortController[] = $state([]);
 
   $effect(() => {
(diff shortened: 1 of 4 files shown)


cintek:
PW-Test: https://github.com/ISDuBA/ISDuBA/actions/runs/26104040883

cintek:
PW-Test: https://github.com/ISDuBA/ISDuBA/actions/runs/26156578485

JanHoefelmeyer (approved):
Works

PR #1506, opened 2026-05-08
fix: Limit in test of query editor

### LINKED ISSUE #1496: Test query can result in large response
I just created a query on an instance with many documents and clicked on "Test query". I got 371087 results but the problem is that the client fetched 169 MB just to get the count because the response also contains the documents.

cintek:
Resolves #1496.


JanHoefelmeyer (approved):
LGTM & works

Issue #1152, opened 2025-10-01
Link or include documentation of advanced search


tschmidtb51:
Currently, one has to know how the advanced filters work to use them. We should either add a link to the current documentation or include the documentation.

JanHoefelmeyer:
This is part of https://github.com/ISDuBA/ISDuBA/issues/95, specifically https://github.com/ISDuBA/ISDuBA/issues/95#issuecomment-2531912979

tschmidtb51:
At least there should be a button that links to the description of search filters.

cintek:
With #1458 we added a link to the documentation.

bernhardreiter:
The link is an improvement, having the current version of the documentation included is even better (as it will match the current software).

bernhardreiter:
(Isn't this done already?)

s-l-teichmann:
@JanHoefelmeyer Why is this still open?

JanHoefelmeyer:
Because it wasn't closed yet, apparently

PR #1516, opened 2026-05-18
Make link to search doc version-specific

cintek:
Before the link to the search documentation always led to the latest documentation but if the user has an older client the documentation might not fit to the software. With this PR we link the documentation that fits to the specific commit which was used to build the client.

This PR implements the suggestion mention in this [comment](https://github.com/ISDuBA/ISDuBA/issues/1152#issuecomment-4460577289).


cintek:
Not necessary anymore because we will serve the documentation directly from ISDuBA itself: #1517.

PR #1517, opened 2026-05-18
Add endpoint to server help for the filter expressions.

s-l-teichmann:
To be serve HTML version of the help of the filter expressions.


JanHoefelmeyer (approved):
LGTM & works

PR #1523, opened 2026-05-20
Fix hover effect of diff indicators

### LINKED ISSUE #1120: Diff selection: Hover color not as expected
In the selection of document versions the color when the cursor hovers over a not-selected button should be light grey. Instead it is blue:

Image

cintek:
Also, use array for classes so it is easier to get an overview

Resolves #1120.


JanHoefelmeyer (approved):
LGTM & works

PR #1533, opened 2026-05-27
Coverage preserved

### LINKED ISSUE #1531: add summary coverage lines to workflow results
The summary lines of the client and backend coverage from `.github/workflows/playwright.yml` should be transfered to the workflow results so they appear below the run under the assets and are preserved as long as the run results are preserved.

Also the tests and coverage should be run on a release, so the coverage can be seen there.

JanHoefelmeyer:
Resolves https://github.com/ISDuBA/ISDuBA/issues/1531

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -29,7 +29,7 @@ jobs:
           ./setup.sh --quick
           ./start_all.sh
           cd ../../client
-          npm run cover
+          npm run cover >> $GITHUB_STEP_SUMMARY
           cd ../
           killall "-2" --wait isdubad
       - name: collect coverage of go code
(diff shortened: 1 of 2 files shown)


.github/workflows/playwright.yml
```diff
@@ -43,7 +45,7 @@ jobs:
           cd go-coverage/$cdir
           # need old style textfmt to be able to calculate total percentage
           go tool covdata textfmt -i=/tmp/cov -o coverage.out
-          go tool cover -func=coverage.out | tee cover_report.txt
+          go tool cover -func=coverage.out | tee cover_report.txt >> $GITHUB_STEP_SUMMARY
```
bernhardreiter: To repeat: We are talking about the summary lines _only_ (Maybe with the table header, but not much more.)

cintek: How does it look now: https://github.com/ISDuBA/ISDuBA/actions/runs/26526608881 ?

bernhardreiter: Much better, I have removed the colors (see commit message for the reason)

bernhardreiter (approved):
LGTM

cintek (approved):
Works

Issue #1409, opened 2026-03-31
Tests and coverage for 0.15.0


JanHoefelmeyer:

cintek:
Coverage in **client** (I will update this comment when there are new numbers):

Before we started to work on 0.15.0 (2026-04-02):

- Bytes: 68.14%
- Statements: 67.15%
- Branches: 47.04%
- Functions: 63.91%
- Lines: 66.91%

New coverage (2026-05-07):

- Bytes: 69.10%
- Statements: 68.70%
- Branches: 48.66%
- Functions: 64.79%
- Lines: 67.60%

bernhardreiter:
Status of coverage of the code base with automatic tests.

https://github.com/ISDuBA/ISDuBA/blob/27040d9954b734c393d0e332a029e39082417479/.github/workflows/playwright.yml
runs some backend and some client tests and the results are partly retained and summarized to an upper level.

Our goal is:
 * run all automatic tests
 * have a summary over the complete code base
 * retain a report (forever at least for the releases)
 * (optional, unrealistic) have an automatic indication how a pull-request would change the coverage

## Backend tests
`BUILD_COVER: "true"` triggers an instrumented build in the Makefile and we get one report about the backend functionality used by the playwright frontend tests.

Seems we are missing the tests from `go test -v ./...` that are in the go.yml workflow.

## Frontend tests
There is https://github.com/ISDuBA/ISDuBA/actions/workflows/unittest.yml which runs `npm run test:unit`. That is included in playwright.yml as the package.json defines:

```json
{
    "cover": "npm run cover:unit && npm run cover:integration",
    "cover:unit": "nyc --silent npm run test:unit"
}
```

bernhardreiter:
#### combined coverage

To get a combined coverage, we would need to be able to sum up the total statements. While the `monocart-report.zip` release asset shows a number of totals (like statements or lines), the `go tool covdata` from https://pkg.go.dev/cmd/covdata does not.

The code here shows that the total statements are calculated for the `percent` output:
https://cs.opensource.google/go/go/+/master:src/internal/coverage/cformat/format.go;drc=40b3c0e58a0ae8dec4684a009bf3806769e0fc41;l=210 , but it is not written out. 

Options:
 * develop or find a modified version of covdata that outputs the total number of statements
 * Use another tool like https://github.com/hhatto/gocloc (LOC) or https://github.com/gengo/goloc (statements) and trust that the statement calculation is the same as in covdata.
[...]

PR #1536, opened 2026-05-28
tests: add more time for long playwright test

bernhardreiter:
Avoid failure because of the variation of how long a test takes to run.

### DIFF
--- a/client/tests/documentView_test.ts
+++ b/client/tests/documentView_test.ts
@@ -21,6 +21,7 @@ test.beforeEach(async ({ page }) => {
 });
 
 test("Advisory view is working", async ({ page }) => {
+  test.slow(); // Easy way to triple the default timeout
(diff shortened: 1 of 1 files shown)


bernhardreiter:
While developing the better go coverage (#1409) I saw a number of playwright tests runs hitting the 30 seconds timeout on this particular test. Raising the timeout to 90 seconds has a better buffer here.

cintek:
@bernhardreiter why not raising the timeout for every test globally (https://playwright.dev/docs/test-configuration#advanced-configuration)? If other tests become longer they might have the same problem.

bernhardreiter:
> why not raising the timeout for every test globally

As I want to watch and notice unexpected behaviour. A unexpected timeout can be a hint that something else broke down.

cintek (approved):
LGTM

PR #1537, opened 2026-05-28
The search.md embedded in the executable needs to be in sync.

s-l-teichmann:
The search.md to be embedded in the executable needs to be in sync with the source.


bernhardreiter (approved):
LGTM

PR #1545, opened 2026-06-16
Resolve issues with URL-manipulation in CVE comparison table

### LINKED ISSUE #1272: cve comparison table issues with URL-manipulation
It's possible to access cve comparison tables of documents without cve or without shared cve by adding /related/documents onto the URL, which shows only a table fragment with a single entry.

Also, manipulating the tracking id while in the related cve page allows writing custom text into the page since the tracking ID is ignored for routing.

cintek:
Resolves #1272.

The solution is consistent to the one that we used when the URL for the advisory view was manipulated.

### DIFF
--- a/client/src/lib/Advisories/Advisory.svelte
+++ b/client/src/lib/Advisories/Advisory.svelte
@@ -31,7 +31,8 @@
     fetchDocumentSSVC,
     fetchSearchHits,
     loadAdvisoryVersions,
-    advisorySearchState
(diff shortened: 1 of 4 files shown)


cintek:
PW-test: https://github.com/ISDuBA/ISDuBA/actions/runs/27624191559

JanHoefelmeyer (approved):
LGTM & works

Issue #1253, opened 2026-01-14
move baseline to versions postgresql-17


bernhardreiter:
0.11.0 postgresql requirement is version >=15.

Postgresql 15 is supported until November 11, 2027 (https://www.postgresql.org/support/versioning/)

In addition newer versions may  have new features that help isduba.

bernhardreiter:
Now that Ubuntu 26.06 LTS is released and Debian Trixie has Postgresql 17, we should do a testing round to adapt to new default postgresql versions in the documentation.

We see significant speed improvements on setups with Debian trixies's Postgresql 17 for example, so it is a strong recommendation to move to it.

bernhardreiter:
We'll add a section on how to migrate an existing database to a newer Postgresql 17 (in general an specific to Ubuntu and Debian). E.g. Using .https://www.postgresql.org/docs/17/pgupgrade.html

PR #1553, opened 2026-07-03
Upgrade doc to real Postgresql version 17

JanHoefelmeyer:
Should be the last thing needed to solve https://github.com/ISDuBA/ISDuBA/issues/1253


4echow (approved):
lgtm + works

PR #1561, opened 2026-07-27
Add Value method to LogLevelType

s-l-teichmann:
Expermiment to  fix strange problem with the upgrade of the PGX driver.


4echow:
fixed via #1570, closing

PR #1570, opened 2026-08-18
Fix invalid input error on pgx 5.10 version bump

4echow:
Fixes the error produced when adding a new feed or updating a feeds log level as described in #1541
by explicitly calling the String method of the FeedLogLevel type variables.

This fix will be necessary once the #1541 bump of pgx to version 5.10 is merged.


s-l-teichmann (approved):
LGTM

PR #1576, opened 2026-08-21
fix: Do not use too many columns in advisory view

### LINKED ISSUE #1262: Layout of advisory view breaks on some documents
Layout for columns too small, for example https://intevation.de/.well-known/csaf/white/2025/intevation-os-2025-0003.json

Image

Tested with 0.11.0 and 0.11.1-31-gde0320ca

cintek:
Resolves #1262.

### DIFF
--- a/client/src/lib/Advisories/Advisory.svelte
+++ b/client/src/lib/Advisories/Advisory.svelte
@@ -771,25 +771,22 @@
           {#if isDiffOpen}
             <Diff showTitle={false}></Diff>
           {:else}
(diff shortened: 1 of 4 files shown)


cintek:
PW-test: https://github.com/ISDuBA/ISDuBA/actions/runs/32469641340

wagner-intevation:
With a 1920px width, before vs after

image 
image

That's WAY better readable.

Also with smaller screen size.

Works well :+1:

JanHoefelmeyer (approved):
LGTM & works

PR #1583, opened 2026-08-25
Make table headers in Search sticky

### LINKED ISSUE #1199: Search: Usability Improvements: Columns unclear
When scrolling down the advisory list in the search function it's not necessary clear which column describes what. Might be confusing when e.g. the order of  cvss2 and cvss3 is no longer clear. Tooltips or scrolling down the headers of the table as well might alleviate the issue.

cintek:
Resolves #1199.

This PR also removes a component that wasn't used and two `div` elements that are not required.


cintek:
PW-test: https://github.com/ISDuBA/ISDuBA/actions/runs/33078396099

JanHoefelmeyer (commented):
LGTM & Works

JanHoefelmeyer (approved):
LGTM & Works

PR #1584, opened 2026-08-26
Improve related docs

### LINKED ISSUE #1282: related CVEs table: sort advisories by count of CVEs
Analysts want to see the advisories first that have more related CVEs
in the "related CVEs table".

The same table should always have the same order of advisories (which are the columns).

Prio: If it can be done with very little work in the client, we can do it in Contract 2025-01,  otherwise it is something for the next contract.

@cintek can you briefly check (and implement if very little work).

### LINKED ISSUE #1579: Limit the number of loaded linked document
### Problem

This issue may improve usability in rare cases where the number of linked documents is very high. For example, SSA-019113 already has 393 linked documents. In this case, the browser is already experiencing performance issues when the link button is accessed.
Image

There is also a case where the number of linked documents exceeded 1,310. In this case, the browser crashed. While this may ultimately depend on the available hardware, displaying such a large number of documents is not useful from a usability perspective either.

### Proposed Solution

-  Reduce the displayed linked documents to those with the highest CVE intersection (i.e., the documents sharing the most CVEs with the current document).
- Set a maximum number of documents to be loaded by default, e.g. 50.
- Allow the user to load all linked documents on demand if required.

cintek:
Resolves #1282 and resolves #1579.

When there are more than 50 related docs a button will be shown to load more at the right side of the table:
Screenshot from 2026-08-31 16-12-30

### DIFF
--- a/client/src/lib/Advisories/RelatedDocuments.svelte
+++ b/client/src/lib/Advisories/RelatedDocuments.svelte
@@ -58,6 +58,7 @@
   let { params = null }: Props = $props();
 
   const uid = $props.id();
+  const STEP = 50;
 
   let document: any | undefined = $state(undefined);
   let documents: RelatedDocument[] | undefined = $state(undefined);
@@ -67,7 +68,7 @@
   let advisoryState: string | undefined = $state(undefined);
   let loadError: ErrorDetails | null = $state(null);
   let isInconsistent = $state(false);
-  let max = $state(50);
+  let max = $state(STEP);
 
   let encodedTrackingID = $derived(
     document?.tracking?.id ? encodeURIComponent(addSlashes(document.tracking?.id)) : undefined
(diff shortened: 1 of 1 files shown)


cintek:
PW-test: https://github.com/ISDuBA/ISDuBA/actions/runs/32986670526

JanHoefelmeyer (dismissed):
Using redhat and bsi as sources.

http://localhost:5173/#/advisories/Red%20Hat%20Product%20Security/RHSA-2024:6122/documents/126/related/documents has currently 121 related advisories.

Clicking Load More... the first 2 times brings in more documents. The third time it does nothing. This is because all 121 related documents are shown, but the button is still shown too. 

This is misleading.

cintek:
PW-test: https://github.com/ISDuBA/ISDuBA/actions/runs/33187521194

cintek:
Since it might be possible that we switch to a virtual grid for the related docs I set this on hold (draft).

PR #1589, opened 2026-08-31
fix: Background color in light theme

cintek:
The PR #1583 led to the issue that the background color was blue at the top of the light theme:

Screenshot from 2026-08-31 20-10-46

This PR solves this issue.


JanHoefelmeyer (approved):
LGTM & works

PR #1590, opened 2026-08-31
Fix cloning queries

### LINKED ISSUE #1564: Not possible to clone single queries
Cloning the (global) relevant queries still works, though.

We should also add a test for this case because until yet we only tested creating, editing and deleting queries but not cloning them.

cintek:
Resolves #1564.


cintek:
PW-test: https://github.com/ISDuBA/ISDuBA/actions/runs/33426556906

JanHoefelmeyer (approved):
LGTM & works

PR #1591, opened 2026-09-01
Using new GoCSAF features  part 1: Use HTTP requests with contexts.

s-l-teichmann:
This PR makes the HTTP requests issued by the server breakable when e.g. the server is going down.


4echow (approved):
lgtm + works

Issue #1593, opened 2026-09-02
context.WithCancel in bulkimport tool


4echow:
Currently the `process` function of the bulkimport tool under `ISDuBA/cmd/main.go` uses a `context.Background` as top level context.
To make the operation cancellable `context.WithCancel` should be used, analogous to the implementation in the isdubad server `run` function under `ISDuBA/cmd/isdubad/main.go`

PR #1596, opened 2026-09-03
Add context cancellation to bulkimport process function

s-l-teichmann:
To make the bulkimport interrupt-able, too.

solves #1593


JanHoefelmeyer (approved):
Works & lgtm

PR #1598, opened 2026-09-03
Add remote validation on temp document import

### LINKED ISSUE #1592: RemoteValidator check for temp documents on import
Currently the `importTempDocument` function under `ISDuBA/pkw/web/tempdocuments.go` only validates the document against the CSAF JSON schema via `csaf.ValidateCSAF` function.

An enhancement would be to use the `csaf.RemoteValidatorWithContext` on the `Controller` receiver to do remote validation tests, if possible.

s-l-teichmann:
resolves #1592


4echow (approved):
lgtm + works

PR #1608, opened 2026-09-09
Fix SQLSTATE error in source feed log count

### LINKED ISSUE #1607: Server error when accessing source feed logs
When accessing the logs of the source feeds, we get a 500 server error, which is produced by an invalid state error 22P02 due to a faulty query we construct in the `ISDUBA/pkg/sources/manager.go` file. 

Image

This is the same error produced by trying to query with the custom `config.FeedLogLevel` type, that no longer implicitly resolves via its `String()` method after the pgx 5.10 version bump, which we fixed for adding new feeds and updating their feed log levels, see #1570 

To fix the behavior we need to explicitly call the `String()` method of the queried log levels.

4echow:
resolves #1607

### DIFF
--- a/pkg/sources/manager.go
+++ b/pkg/sources/manager.go
@@ -818,7 +818,7 @@ func (m *Manager) StreamFeedLog(
 				cond.WriteString(` OR `)
 			}
 			fmt.Fprintf(&cond, "lvl = $%d", len(args)+1)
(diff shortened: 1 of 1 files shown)


JanHoefelmeyer (approved):
LGTM & works

